PR #16, opened 2021-05-14
Support doc generation via Makefile

liorr:
> JIRA Ticket URL: https://twingate.atlassian.net/browse/TG-2994

## Changes
- `make docs` now uses `GOBINPATH` to run `tfplugindocs`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,19 +9,20 @@ GOBINPATH=$(shell go env GOPATH)/bin
 
 default: build
 
-.PHONY: docs
-
 .PHONY: ci-checks
 ci-checks: docs
 	echo "Checking if latest docs generated"
 	git diff --exit-code || echo "ERROR: Update and push the latest documentation"; exit 1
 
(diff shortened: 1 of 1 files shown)


romankor (approved):
LGTM , but i am a little confused by the `.PHONY` targets , its is a bit pointless to define all targets `phony` , 
As much as i understand `phony` targets are just defined to resolve clashes between real file names in the directory

liorr:
> M , but i am a little confused by the `.PHONY` targets , its is a bit pointless to define all targets `phony` ,
> As much as i understand `phony` targets are just defined to resolve clashes between real file names in the directory

Most of the `.PHONY`s were at the bottom, I just think it make sense to put them next to the target as it clearly shows which one has it and which one doesn't. .
`There are two reasons to use a phony target: to avoid a conflict with a file of the same name, and to improve performance.`
https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

PR #28, opened 2021-06-17
TG-2704 | isActive parameter is unnecessary and being deprecated

ekampf:
> JIRA Ticket URL: TG-2704 

## Changes
- When creating a remote network it is not necessary to send an `isActive` argument


romankor (approved):
GTM

PR #29, opened 2021-06-18
TG-3273 | fix resource order

ekampf:
> JIRA Ticket URL: TG-3273

## Changes
- Rresource groups relationship should be treated as a set not an ordered list (order doesn't matter)


romankor (approved):
GTM

PR #74, opened 2022-04-15
Updated graphql client in provider to expose string types instead of graphql

vmanilo:
## Changes
- Changed internal client uses in provider, so from terraform provider we don't need to use graphql types


romankor (approved):
LGTM

Issue #77, opened 2022-04-19
When supplying an empty port range with the "RESTRICTED" policy option in the twingate_resource.protocols resource definition, the provider crashes


alexmensch:
Expected behavior: this configuration is valid and the provider should not crash. Per our public API behavior, when supplying an empty port list in the "RESTRICTED" policy mode, traffic on all ports will be blocked.

Let me know if you need reproduction steps @vmanilo

vmanilo:
the issue is clear, what would be the expected output?
to return error msg `empty port range`?

alexmensch:
Ah yeah, sorry, I'll edit the original description. But this is actually valid, so it should be accepted without crashing and simply block all ports on the relevant protocol.

When an empty port range is supplied in our API with the "RESTRICTED" policy option, this means that all ports are blocked. This is similar to the "DENY_ALL" issue, which is just a shortcut for this same combination of "RESTRICTED" with an empty port range.

vmanilo:
@alexmensch I've manage to reproduce crash only with configuration like this, did you mean this case?

```
tcp  {
      policy = "RESTRICTED"
      ports = [""]
    }
```

vmanilo:
and also this crashes 

```
ports = ["", "80"]
```

so, basically if we have at least one empty string, we can ignore all other ports in the list, right?

vmanilo:
btw, when passing values like this

```
      policy = "RESTRICTED"
      ports = [" "]
```

backend converts it to "ALLOW_ALL" (see this value when fetching resource)

vmanilo:
should I keep policy name **RESTRICTED**, or I need to change it to **DENY_ALL**?

ekampf:
@vmanilo no it doesn't need to change.
You should look at the GraphQL docs for `ProtocolsInput` (and maybe they're not clear enough and we need to fix them?):
![Screen Shot 2022-04-22 at 1 01 23 PM](https://user-images.githubusercontent.com/205185/164785699-9cc761c0-7adc-43aa-ae8c-cb2599fa7b50.png)

So `ProtocolsInput` has 2 "modes":
1. `ALLOW_ALL` in which case all ports are allowed and we dont even look at the `ports` value
2. `RESTRICTED` in which only traffic to certain ports is allowed, per the definition in the `ports` field.

If a customer wants to restrict all access to a protocol he would need to specify:
```
policy = "RESTRICTED"
ports = []
```

[...]

PR #78, opened 2022-04-23
Fixed crashing on empty port range

vmanilo:
Github Issue: https://github.com/Twingate/terraform-provider-twingate/issues/77

## Changes
- fixed crashing when user passes empty string to ports range
- added tests


twingate/client_resource.go
```diff
@@ -124,6 +128,12 @@ func convertPortsToSlice(a []interface{}) []string {
 	var res = make([]string, 0)
 
 	for _, elem := range a {
+		if elem == nil {
+			res = append(res, "")
+
+			continue
+		}
```
vmanilo, line 135: @romankor here is the fix

twingate/resource_resource_test.go
```diff
@@ -165,7 +165,124 @@ func testTwingateResource_fullFlowCreation(networkName, resourceName string) str
...
+	return fmt.Sprintf(`
+	resource "twingate_remote_network" "test" {
+	  name = "%s"
+	}
+
+	resource "twingate_resource" "test" {
+	  name = "%s"
+	  address = "updated-acc-test.com"
+	  remote_network_id = twingate_remote_network.test.id
+	  group_ids = ["R3JvdXA6MjMxNTQ="]
```
romankor, line 201: lets use `resource_group` to create this group and use it in the the test

twingate/resource_resource_test.go
```diff
@@ -165,7 +165,124 @@ func testTwingateResource_fullFlowCreation(networkName, resourceName string) str
...
```
romankor, line 272: ttl

romankor (changes requested):
See comments

twingate/resource_resource_test.go
vmanilo: added resource_group

twingate/resource_resource_test.go
vmanilo: added resource_group

romankor (approved):
LGTM

PR #80, opened 2022-04-28
Ensure the root go.mod reflects a correct go module structure

stack72:
## Changes
- Make the Go module importable


emrul:
Context: The reason for the PR is so that in the Pulumi provider you can import the Go module.

The Pulumi provider wraps our TF provider so the Pulumi code wants to import it and needs an absolute reference to it.

vmanilo:
@emrul as far as I know to be able use terraform-bridge for pulumi provider, our terraform provider should use plugin-sdk v1, and currently we using v2, I guess that can make an issue in integration

stack72:
Hi @vmanilo 

No we support v2 of the plugin SDK :)

Paul

ekampf (approved):

Issue #94, opened 2022-05-19
When importing a connector the `remote_network_id` is not stores in the state 


romankor:
When importing an existing connector its `remote_network_id` is not saved in the state , so next time plan is run it will try to recreate it 

this is a properly created connector :

```
terraform apply
twingate_remote_network.aws_network: Refreshing state... [id=UmVtb3RlTmV0d29yazozODg0OA==]

Terraform used the selected providers to generate the following execution plan. Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  # twingate_connector.aws_connector will be created
  + resource "twingate_connector" "aws_connector" {
      + id                = (known after apply)
      + name              = (known after apply)
      + remote_network_id = "UmVtb3RlTmV0d29yazozODg0OA=="
    }

Plan: 1 to add, 0 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

twingate_connector.aws_connector: Creating...
twingate_connector.aws_connector: Creation complete after 0s [id=Q29ubmVjdG9yOjI5OTA=]

Apply complete! Resources: 1 added, 0 changed, 0 destroyed.

➜ terraform state  show twingate_connector.aws_connector
# twingate_connector.aws_connector:
resource "twingate_connector" "aws_connector" {
    id                = "Q29ubmVjdG9yOjI5OTA="
    name              = "garnet-elk"
    remote_network_id = "UmVtb3RlTmV0d29yazozODg0OA=="
}
```

this is an imported one: 

```
terraform import twingate_connector.aws_connector Q29ubmVjdG9yOjE4OTQ=
twingate_connector.aws_connector: Importing from ID "Q29ubmVjdG9yOjE4OTQ="...
twingate_connector.aws_connector: Import prepared!
  Prepared twingate_connector for import
twingate_connector.aws_connector: Refreshing state... [id=Q29ubmVjdG9yOjE4OTQ=]

Import successful!

The resources that were imported are shown above. These resources are now in
your Terraform state and will henceforth be managed by Terraform.

terraform state  show twingate_connector.aws_connector
# twingate_connector.aws_connector:
resource "twingate_connector" "aws_connector" {
    id   = "Q29ubmVjdG9yOjE4OTQ="
    name = "piquant-manul"
}

terraform plan
[...]

PR #96, opened 2022-05-21
Fix importing a connector

vmanilo:
Github Issue: https://github.com/Twingate/terraform-provider-twingate/issues/94

## Changes
- Fixed connector import - set network_id

### DIFF
--- a/docs/resources/connector.md
+++ b/docs/resources/connector.md
@@ -47,5 +47,6 @@ resource "twingate_connector" "aws_connector" {
 Import is supported using the following syntax:
 
 ```shell
+terraform import twingate_remote_network.aws_network UmVtb3RlTmV0d29yazozOTkwNw==
 terraform import twingate_connector.connector Q29ubmVjdG1b0qe0
 ```
--- a/docs/resources/connector.md
+++ b/docs/resources/connector.md
@@ -47,6 +47,5 @@ resource "twingate_connector" "aws_connector" {
 Import is supported using the following syntax:
 
 ```shell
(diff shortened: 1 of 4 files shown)


docs/resources/connector.md
```diff
@@ -47,5 +47,6 @@ resource "twingate_connector" "aws_connector" {
 Import is supported using the following syntax:
 
 ```shell
+terraform import twingate_remote_network.aws_network UmVtb3RlTmV0d29yazozOTkwNw==
```
romankor, line 50: i don't think we need to add these imports to documentation, its kinda obvious that a network has to be present in order for the connector to work

examples/resources/twingate_connector/import.sh
```diff
@@ -1 +1,2 @@
-terraform import twingate_connector.connector Q29ubmVjdG1b0qe0
\ No newline at end of file
+terraform import twingate_remote_network.aws_network UmVtb3RlTmV0d29yazozOTkwNw==
```
romankor, line 1: same here

romankor (changes requested):
Fix .. remove imports from documentation, but I verified the fix and it works

docs/resources/connector.md
vmanilo: removed it

examples/resources/twingate_connector/import.sh
vmanilo: removed it

romankor (approved):
GTM

Issue #64, opened 2022-04-07
Add data source: twingate_remote_network


bertekintw:
Arguments

- (Required) `id`
    - This ID can be obtained either directly from the public admin API or referenced via the `twingate_remote_networks` data source in our Provider.

Exported attributes

- `id` (String)
- `name` (String)

GraphQL

```graphql
query {
  remoteNetwork(id:"UmVtb3RlTmV0d29yazoyMg==")
   {
    id
		name
  }
}
```

alexmensch:
> hey @alexmensch can you check description, when we fetching remote-network by id in result I'd expected single object - not a list, and description says - A list of remote_networks

Yeah, you're absolutely right. Let me see what happened—I think it might have been a copy & paste error across all the different issue descriptions—but I'll update the description to make sure that the PR ultimately matches the issue.

alexmensch:
Updated @vmanilo!

vmanilo:
@alexmensch please update description according to new plans

alexmensch:
In closing #103 due to the controller change, we still need a way to find a remote network based on its name, which is the most important part of the new data sources. I'm closing this issue with the release of provider v1.0.8, and opening up a different issue to add a name lookup for a single remote network.

PR #112, opened 2022-06-06
Remote network datasource

vmanilo:
Github Issue: https://github.com/Twingate/terraform-provider-twingate/issues/64

## Changes
- added datasource for remote-network


twingate/datasource_remote_network.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	resourceData.SetId(networkID)
+
+	return diags
+}
+
+func datasourceRemoteNetwork() *schema.Resource {
+	return &schema.Resource{
+		Description: "A Remote Network represents a single private network in Twingate that can have one or more Connectors and Resources assigned to it. You must create a Remote Network before creating Resources and Connectors that belong to it. For more information, see Twingate's [documentation](https://docs.twingate.com/docs/remote-networks).",
```
vmanilo, line 33: @alexmensch please check namings

twingate/datasource_remote_network.go
```diff
@@ -0,0 +1,48 @@
...
+		ReadContext: datasourceRemoteNetworkRead,
+		Schema: map[string]*schema.Schema{
+			"id": {
+				Type:        schema.TypeString,
+				Required:    true,
+				Description: "The ID of the Remote Network",
```
vmanilo, line 39: @alexmensch please check namings

twingate/datasource_remote_network.go
```diff
@@ -0,0 +1,48 @@
...
+			},
+			"name": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The name of the Remote Network",
```
vmanilo, line 44: @alexmensch please check namings

examples/data-sources/twingate_remote_network/data-source.tf
```diff
@@ -0,0 +1,17 @@
+provider "twingate" {}
+
+resource "twingate_remote_network" "net1" {
```
ekampf, line 3: Looking at docs for others datasources (like [here](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/pubsub_topic) and [here](https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/data_source_dataproc_metastore_service)) they dont include the resource creation in the example.

It makes sense because the example doesn't make sense - if you create it via TF you dont need to reference it by name via datasource.

Any reason this example shouldn't just be:
```
data "twingate_remote_network" "foo" {
   id = "<your network's id>"
 }
```

alexmensch (approved):
All the text looks great to me, thank you!

examples/data-sources/twingate_remote_network/data-source.tf
vmanilo: sure, we can do like that, 
my intention was to provide working example, but your proposal also make sense

will update it

vmanilo: @ekampf updated example

romankor (approved):
LGTM

alexmensch:
@ekampf just waiting on your review and then we're good to go with this PR

ekampf (approved):

Issue #99, opened 2022-05-25
Add data source: twingate_groups


alexmensch:
Arguments

- (Optional) `name`
    - Filters the list of groups returned by an exact name match (case sensitive).
    - Note that group names are not required to be unique, so multiple groups may be returned.

Exported attributes

- A list of `groups` with each element containing the following information:
    - `id` (String)
    - `name` (String)

GraphQL

```graphql
query {
  groups(filter: { name: { eq: "Admins" } } )
   {
    edges {
      node {
        id
		    name
      }
    }
  }
}
```

tjstansell:
Oh, and I would argue that this should _always_ return a list, even if that list is empty.  As in, it should never error.

alexmensch:
> Similar to my comments about the `twingate_group` (singular) resource, ideally, this should return a list of groups that match a series of filters to help narrow down the results. Perhaps you'd want a list of all okta-synced groups. You might do
> 
> ```terraform
> resource "twingate_groups" "okta" {
>   type     = "SYNCED"
>   isActive = true
> }
> ```
> 
> Basically, this resource should be as flexible as possible to let folks target the groups they want.

Makes sense -- we'll be tracking this in issues #137 and #138.

alexmensch:
> Oh, and I would argue that this should _always_ return a list, even if that list is empty. As in, it should never error.

@vmanilo what is the current implementation behavior if the data source returns no groups? Does it error or return an empty list?

vmanilo:
@alexmensch it will return error

alexmensch:
> @alexmensch it will return error

Ok, agree with @tjstansell that this should return an empty list if the name argument doesn't match any groups. The behavior would otherwise be a bit odd if filtering for a group that doesn't exist breaks your Terraform run. Do you see any reasons @vmanilo or @ekampf why we wouldn't want to return an empty list and erroring out would be preferred? If another part of your Terraform configuration depends on the list being non-empty, you'll get an error at that point anyway.

alexmensch:
Great, we'll go ahead and make this change. Re-opening this issue @vmanilo

PR #117, opened 2022-06-14
Groups datasource (plural)

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/99

## Changes
- added datasource for groups (plural)
- moved client errors in separate file


twingate/datasource_groups.go
```diff
@@ -0,0 +1,70 @@
...
+					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: "The ID of the Group",
+						},
+						"name": {
+							Type:        schema.TypeString,
+							Computed:    true,
+							Description: "The name of the Group",
```
vmanilo, line 63: @alexmensch please check descriptions

twingate/datasource_groups.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func datasourceGroups() *schema.Resource {
+	return &schema.Resource{
+		Description: "Groups are how users are authorized to access Resources. For more information, see Twingate's [documentation](https://docs.twingate.com/docs/groups).",
+		ReadContext: datasourceGroupsRead,
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:        schema.TypeString,
+				Required:    true,
```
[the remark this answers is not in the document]
vmanilo: ok, will update PR

docs/data-sources/groups.md
```diff
@@ -0,0 +1,41 @@
...
+Groups are how users are authorized to access Resources. For more information, see Twingate's [documentation](https://docs.twingate.com/docs/groups).
+
+## Example Usage
+
+```terraform
+data "twingate_groups" "foo" {
+  name = "<your group's name>"
+}
+```
+
```
alexmensch, line 20: Let's add a clarification comment below this to give some context as to why this is a plural data source.

`Group names are not constrained to be unique within Twingate, so it is possible that this data source will return multiple list items.`

alexmensch (changes requested):

docs/data-sources/groups.md
vmanilo: updated docs

alexmensch: Ah, I see that this is within the Terraform example itself, and now I understand the `#` comment notation. Originally I was thinking that we should change this to move it up into the general description, but I think this works. Will approve this and the other similar changes and then we can merge everything.

alexmensch (approved):

Issue #130, opened 2022-06-27
Terraform does not detect a disabled state change and therefore does not re-enable the service


agriffin208:
Any changes to Terraform resources that can have a state changed (disabled) from the Twingate Admin Console should be detected. Relates to #79 when running v0.1.7.

1. Create a twingate_resource.
2. Disable resource from Twingate Admin Console
3. Apply TF
4. Check Resources state: Resource state is not changed
5. Delete the Resource from Twingate Admin Console
6. Apply TF
7. Error:
`Refreshing state... [id=<id>]
╷
│ Error: failed to read resource with id <id>
│
│ with twingate_resource.<name>,
│ on test.tf line 103, in resource "twingate_resource" "<name>":
│ 103: resource “twingate_resource"`

PR #131, opened 2022-06-28
Fix set active resource state on terraform apply

vmanilo:
Fixes https://github.com/Twingate/terraform-provider-twingate/issues/130

PR addresses first part of the issue

## Changes
- fixing issue with setting active state to resource when running terraform apply


twingate/resource_resource.go
```diff
@@ -322,6 +322,12 @@ func resourceResourceRead(ctx context.Context, resourceData *schema.ResourceData
 		return diag.FromErr(err)
 	}
 
+	// fix active resource state on terraform apply
+	_ = client.updateResourceActiveState(ctx, &Resource{
```
ekampf, line 326: Not sure I understand this fix.
Whenever you read a resource you also issue a mutation call to set it to be active? 
Why not check if its already active first?

vmanilo: thank you, updated it

ekampf (approved):

PR #132, opened 2022-06-29
Set Go version to 1.18.2

ekampf:
`go.mod` was set to `1.16` but we were running CI with `1.18.2`


.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ jobs:
     environment:
       TEST_RESULTS: /tmp/test-results
     docker:
-      - image: cimg/go:1.18.2
+      - image: cimg/go:1.17.11
```
vmanilo, line 9: we need to do the same changes for build stage (line 49)

```
  build:
    docker:
      - image: cimg/go:1.18.2
```

vmanilo: and what about github workflow https://github.com/Twingate/terraform-provider-twingate/blob/main/.github/workflows/release.yml#L31

should we update here as well?

vmanilo: should we update in README file https://github.com/Twingate/terraform-provider-twingate/blob/main/README.md?plain=1#L11 as well?

PR #135, opened 2022-06-29
Fix resource re-creation after deletion

vmanilo:
Resolves (github issue): https://github.com/Twingate/terraform-provider-twingate/issues/130 (second part)

## Changes
- fix resource re-creation after deletion


ekampf (changes requested):
@vmanilo we need the same fix for all other resources too (group, remote network, connector...)
Same use case applies to all of them - someone manually deleted them creating a drift that TF needs to fix

romankor (approved):
I compiled that version of the provider locally and tried to create and then delete the resources in the background and  TF always recreates them with no issues. 
So LGTM here

Issue #139, opened 2022-06-29
`twingate_groups` datasource - the `name` parameter should be optional


tjstansell:
You should be able to get a list of all groups.  #138 would let us optimize this query in the future and let the API filter the results on the server side. #137 will let us get more data to find the group we're looking for based on more attributes which would let us filter things ourselves.  Even without those, we should be able to just query for all groups and get a list of `id` and `name`s.  This, at least, would be a single api call instead of one per group you might be interested in.  And if you know you don't have multiple groups of the same name, you can pretty easily find the `id` of the group you're interested in.

alexmensch:
Hey @tjstansell the intent in #99 was to make the `name` parameter optional. @vmanilo can you confirm that that's the case? If not, we need to make sure we fix that.

Issue #137, opened 2022-06-29
Expose additional read-only attributes for twingate_group and twingate_groups data sources


alexmensch:
### Summary

Exposing additional attributes would be helpful in order to differentiate between groups on more information than the name of the group.

The following attributes are currently available from the public Admin API:
- isActive (Boolean)
- type (Enum) [ MANUAL, SYNCED, SYSTEM ]

### Additional exported attributes

- `is_active` (Boolean)
- `type` (String)
- - The set of `type` attribute values must be `[ MANUAL, SYNCED, SYSTEM ]`, corresponding to the public API enum values

### Admin API GraphQL query

query {
  group(id: "R3JvdXA6MjM0")
   {
    id
    name
    isActive
    type
  }
}

tjstansell:
I think that if this issue and #139 are resolved the `twingate_groups` datasource becomes functional for a majority of use-cases.  It requires filtering within terraform, but that should be fine.  Future updates (e.g. #138) would be additional optimizations to reduce the number of API calls or reduce the payload from those API calls, which is great, but not a priority, IMO.

alexmensch:
> I think that if this issue and #139 are resolved the `twingate_groups` datasource becomes functional for a majority of use-cases. It requires filtering within terraform, but that should be fine. Future updates (e.g. #138) would be additional optimizations to reduce the number of API calls or reduce the payload from those API calls, which is great, but not a priority, IMO.

Perfect, yes, we're on the same page here @tjstansell. Appreciate the effort you're putting into the feedback here.

PR #143, opened 2022-07-03
Added optional filter params for groups datasource

vmanilo:
Resolves (github issue):
https://github.com/Twingate/terraform-provider-twingate/issues/137
https://github.com/Twingate/terraform-provider-twingate/issues/99
https://github.com/Twingate/terraform-provider-twingate/issues/139

## Changes
- fixed groups datasource


twingate/datasource_groups.go
```diff
@@ -22,30 +28,74 @@ func datasourceGroupsRead(ctx context.Context, resourceData *schema.ResourceData
...
+			"is_active": bool(group.IsActive),
 		})
 	}
 
 	if err := resourceData.Set("groups", data); err != nil {
 		return diag.FromErr(err)
 	}
 
-	resourceData.SetId(groupName)
+	resourceData.SetId("groups-by-filter")
```
tjstansell, line 42: I don't know enough about how datasources work, but does setting this to a static value:
* cause refresh problems?
* cause problems with multiple data sources?

docs/data-sources/groups.md
```diff
@@ -24,9 +24,11 @@ data "twingate_groups" "foo" {
 <!-- schema generated by tfplugindocs -->
 ## Schema
 
-### Required
+### Optional
 
+- `is_active` (Boolean) Indicates if the Group is active
```
tjstansell, line 29: Personally, I would use descriptions that indicate exactly how these are used to filter.
```
- `is_active` (Boolean) Returns only groups matching the specified state.
- `name` (String) Returns only groups that exactly match this name.
- `type` (String) Returns only groups of the specified type (valid: `MANUAL`, `SYNCED`, `SYSTEM`).
```

twingate/client_group.go
```diff
@@ -210,3 +230,65 @@ func (client *Client) deleteGroup(ctx context.Context, groupID graphql.ID) error
...
+
+	if f.IsActive != nil && *f.IsActive != bool(group.IsActive) {
+		return false
+	}
+
+	return true
+}
+
+func (client *Client) filterGroups(ctx context.Context, filter *GroupsFilter) ([]*Group, error) {
+	if filter == nil {
```
tjstansell, line 253: If `buildFilter()` always returned a valid `GroupsFilter` object (even if all the attributes are `nil`), you could remove this whole block to simplify the code a bit.  You're already optimizing below which query you want to make to the API based on the individual filters, so this block is essentially a duplicate of all that.

vmanilo: thank you, good point

twingate/datasource_groups.go
vmanilo: tested locally -  all good

also added test with two datasource

docs/data-sources/groups.md
vmanilo: @alexmensch what do you think?

twingate/client_group.go
vmanilo: fixed

docs/data-sources/groups.md
alexmensch: Yes agreed -- @vmanilo go ahead as above, just need to capitalize groups to `Groups`. Thanks all!

alexmensch (changes requested):

docs/data-sources/groups.md
vmanilo: fixed

alexmensch (approved):

twingate/datasource_groups.go
```diff
@@ -22,29 +29,73 @@ func datasourceGroupsRead(ctx context.Context, resourceData *schema.ResourceData
...
 
+func buildFilter(resourceData *schema.ResourceData) *GroupsFilter {
+	groupName, hasName := resourceData.GetOk("name")
+	groupType, hasType := resourceData.GetOk("type")
+
+	// GetOk does not provide correct value for exists flag (second output value)
+	groupIsActive, hasIsActive := resourceData.GetOkExists("is_active") // nolint:staticcheck
+
+	if !hasName && !hasType && !hasIsActive {
+		return nil
```
tjstansell, line 56: How is this not causing problems when there is no filtering and you pull all groups?

vmanilo: I'm handling nil case in the GroupsFilter methods

twingate/client_group.go
```diff
@@ -214,3 +233,60 @@ func (client *Client) deleteGroup(ctx context.Context, groupID graphql.ID) error
...
+			return nil, nil
+		}
+
+		return nil, err
+	}
+
+	var filtered []*Group
+
+	for _, g := range groups {
+		if filter.Match(g) {
```
ekampf, line 286: @alexmensch did we agree not to allow filtering by `name` until server api supports it?

twingate/datasource_groups_test.go
```diff
@@ -84,19 +86,161 @@ func testTwingateGroupsDoesNotExists(name string) string {
...
+		resource.Test(t, resource.TestCase{
+			ProviderFactories: testAccProviderFactories,
+			PreCheck:          func() { testAccPreCheck(t) },
+			CheckDestroy:      testAccCheckTwingateGroupDestroy,
+			Steps: []resource.TestStep{
+				{
+					Config: testDatasourceTwingateGroupsWithDatasource(groupName),
+					Check: resource.ComposeTestCheckFunc(
+						resource.TestCheckOutput("my_group", groupName),
+						resource.TestCheckResourceAttr("data.twingate_groups.out", "groups.#", "2"),
```
ekampf, line 205: Fix test - if both DS have the same value how do we know they're not overriding each other?

vmanilo: updated test

twingate/client_group.go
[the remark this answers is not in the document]
alexmensch: Yes that's correct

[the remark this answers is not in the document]
alexmensch: But if it's implemented already...may as well ship it? What do you think @ekampf?

Issue #144, opened 2022-07-05
Ensure that we return an empty list for all plural data sources


alexmensch:
Per comment in issue #99

PR #147, opened 2022-07-06
Fix empty result for resources datasource

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/144

## Changes
- fix for resources datasource when we get empty result


ekampf (changes requested):
Not sure I understand this fix.
`Edges` will never be `nil` is a required field per our schema.
It can only be `nil` in an error edge case (query in invalid or unknown server error) 


Screen Shot 2022-07-06 at 11 38 31 PM

twingate/client_resource.go
```diff
@@ -410,8 +410,8 @@ func (client *Client) readResourcesByName(ctx context.Context, name string) ([]*
 		return nil, NewAPIErrorWithID(err, "read", resourceResourceName, "All")
 	}
 
-	if len(response.Resources.Edges) == 0 {
-		return nil, NewAPIErrorWithID(ErrGraphqlResourceNotFound, "read", resourceResourceName, "All")
+	if response.Resources.Edges == nil {
```
ekampf, line 413: This is wrong. `Edges` can't be `nil`.
The previous statement was correct its just that it shouldnt raise an error but return an empty list result

ekampf (approved):

PR #154, opened 2022-07-20
Remove reference to requiring Enterprise tier for API key

emrul:
Prospect thought they needed Enterprise Tier to use Terraform and referenced this doc. Since we changed our tiers this is no longer correct so submitting a PR for this.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -13,7 +13,7 @@ The Twingate provider can be used with [Twingate](https://www.twingate.com) to m
 
 ## Twingate Setup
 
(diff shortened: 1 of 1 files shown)


alexmensch (approved):

ekampf:
@emrul @alexmensch you guys fixed the wrong file - everything in `docs/` is autogenerated so this change will be overriden the next time someone `make docs` (I;m adding a CI step to fail CI if someone makes this mistake)

The file you need to edit  is at `/templates/indesx.md.tmpl`

PR #155, opened 2022-07-20
Move CI to Github Actions

ekampf:
Based on build process from Hashicorp's https://github.com/hashicorp/terraform-provider-scaffolding

# Changes
- Move CI from CircleCI to Github
- Run acceptance tests on multiple TF versions to ensure compatibility
- Coverage now uploaded for all tests not just acceptance (it was a bug in the CircleCI build)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,155 @@
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v3
+        with:
+          go-version-file: 'go.mod'
+          cache: true
+      - run: go generate ./...
+      - name: git diff
+        run: |
+          git diff --compact-summary --exit-code || \
+            (echo; echo "Unexpected difference in directories after code generation. Run 'go generate ./...' command and commit."; exit 1)
```
vmanilo, line 51: I think we don't use code generation with `go generate` cmd

but we could use this job for checking if latest docs was generated, so we could run `make docs` instead

ekampf: Fixed!

romankor (approved):
LGTM

PR #166, opened 2022-08-03
Add `TEST_UNIQUE_VALUE` to be used to isolate ACC Tests runs

ekampf:
Initial work on #167 

## Changes
- Adds `TEST_UNIQUE_VALUE` env var to be used by tests to isolate themselves


liorr (approved):
GTM

PR #170, opened 2022-08-05
added cleanup to smoke tests; disabled environment

vmanilo:
## Changes
- test ci

PR #171, opened 2022-08-12
CI: Dont run acceptance tests from forks

ekampf:
# Changeas
- Downgraded `lint` as new version has breaking changes. We already have a separate PR to upgrade it - https://github.com/Twingate/terraform-provider-twingate/pull/170
- Removed acceptance-tests from forked PRs
- Added concurrency settings so that pushing a change cancels previous running build
-


liorr (approved):
GTM

PR #172, opened 2022-08-13
CI: Cache go dependencies

ekampf:
Adding cache to go dependencies to reduce random failures due to networking errors on `go mod download`


romankor (approved):
GTM

PR #181, opened 2022-08-22
Refactoring: Restructure project layout

vmanilo:
## Changes: set new project layout
- twingate
  - provider.go                  --  main provider file
  - internal
    - client                          -- implements graphql queries to the backend
    - model                         -- implements data objects which consumes provider code 
    - provider
      - datasource              -- provider datasources 
      - resource                  -- provider resources 
    - test                             --  main package with tests
      - acctests
        - datasource
        - resource
      - client
      - sweepers
    - utils

Now provider code doesn't depend on graphql library, it uses only in the client package and provider operates with model objects


docs/index.md
```diff
@@ -36,7 +36,7 @@ provider "twingate" {
 - `api_token` (String, Sensitive) The access key for API operations. You can retrieve this
 from the Twingate Admin Console ([documentation](https://docs.twingate.com/docs/api-overview)).
 Alternatively, this can be specified using the TWINGATE_API_TOKEN environment variable.
-- `http_max_retry` (Number) Specifies a retry limit for the http requests made. This default value is 10.
+- `http_max_retry` (Number) Specifies a retry limit for the http requests made. This setting is 10.
```
ekampf, line 39: Why did this change?   the correct working is the original ("The default value is")

twingate/provider.go
```diff
@@ -46,42 +58,42 @@ func providerOptions() map[string]*schema.Schema {
...
+				"Alternatively, this can be specified using the %s environment variable", DefaultHTTPTimeout, EnvHTTPTimeout),
 		},
 		"http_max_retry": {
 			Type:        schema.TypeInt,
 			Optional:    true,
-			DefaultFunc: schema.EnvDefaultFunc("TWINGATE_HTTP_MAX_RETRY", DefaultHTTPMaxRetry),
-			Description: "Specifies a retry limit for the http requests made. This default value is 10.\n" +
-				"Alternatively, this can be specified using the TWINGATE_HTTP_MAX_RETRY environment variable",
+			DefaultFunc: schema.EnvDefaultFunc(EnvHTTPMaxRetry, DefaultHTTPMaxRetry),
+			Description: fmt.Sprintf("Specifies a retry limit for the http requests made. This setting is %s.\n"+
```
ekampf, line 95: This shouldn't have changed

docs/index.md
vmanilo: thank you for the comment, yeah really not sure at which point I've change it, but agree with your proposal.
Will change it to original one

twingate/provider.go
vmanilo: right, thanks for the fix

docs/index.md
vmanilo: regenerated docs

PR #190, opened 2022-09-15
Try to fix random failure

vmanilo:
We have issue with not stable acceptance tests, which periodically randomly fails, this PR is a follow up attempt to improve and stabilise tests.

## Changes
- updated acc test

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,21 +5,21 @@ name: CI
 permissions: read-all
 on:
   pull_request:
-#    branches:
-#      - main
+    branches:
+      - main
     paths-ignore:
        - 'README.md'
 
   push:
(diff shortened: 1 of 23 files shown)


ekampf (changes requested):
You added a second wait between any call to server. 
Its incorrect.

The only issue we have is when a resource is read immediately after its being written.
We need to find those cases and fix those, and not just add a second delay before any network operation.

vmanilo:
@ekampf I've addressed the issue you mentioned, checked matrix tests on my fork - looks good

ekampf (approved):
Lets reduce `waitForResourceAvailability()` from `1s` to half a second
Fix and merge

ekampf:
Lets also change integration tests to run every 10m so we see they're stable

PR #194, opened 2022-10-04
Fix connector resource read/create/update operations

vmanilo:
Update connector resource

## Changes
- fix read/create/update operations
- extracted connector tokens


twingate/resource_connector.go
```diff
@@ -129,15 +115,22 @@ func resourceConnectorRead(ctx context.Context, resourceData *schema.ResourceDat
 		return diag.FromErr(err)
 	}
 
-	if err := resourceData.Set("name", connector.Name); err != nil {
+	if err := resourceData.Set("name", string(connector.Name)); err != nil {
 		return diag.FromErr(fmt.Errorf("error setting name: %w ", err))
 	}
 
+	var connectorRemoteNetworkID string
 	if connector.RemoteNetwork != nil {
```
ekampf, line 123: Is this check necessary? 
`remoteNetwork` is a required field on Connector. There is no way you reached here and it's `nil` because you wouldnt have gotten an `err` if server returned such response
Screen Shot 2022-10-14 at 5 15 53 PM

vmanilo: I see your point, yes we can simplify this code

vmanilo: fixed

ekampf (approved):

PR #196, opened 2022-10-10
Fix resource create/update

vmanilo:
Resolves  resource create/update operation

## Changes
- added read helper method


twingate/client_resource.go
```diff
@@ -21,6 +21,46 @@ type Resource struct {
...
+	}
+	RemoteNetwork struct {
+		ID graphql.ID
+	}
+	Groups struct {
+		PageInfo struct {
+			HasNextPage graphql.Boolean
+		}
+		Edges []*Edges
+	} `graphql:"groups(first: $first)"`
```
ekampf, line 38: `$first` is not a good name as its very generic - what if we had multiple different connections we wanted to paginate here?
This should be something like `$groupsFirst` or `$groupsPageSize`

twingate/client_resource.go
```diff
@@ -145,56 +185,40 @@ func convertPorts(ports []string) ([]*PortRangeInput, error) {
...
 
-func (client *Client) createResource(ctx context.Context, resource *Resource) error {
+func (client *Client) createResource(ctx context.Context, resource *Resource) (*Resource, error) {
 	variables := map[string]interface{}{
 		"name":            resource.Name,
 		"address":         resource.Address,
 		"remoteNetworkId": resource.RemoteNetworkID,
 		"groupIds":        resource.GroupsIds,
 		"protocols":       resource.Protocols,
+		"first":           graphql.Int(readResourceQueryGroupsSize),
```
ekampf, line 199: See comment above - https://github.com/Twingate/terraform-provider-twingate/pull/196/files#r996166727

twingate/client_resource.go
```diff
@@ -259,32 +267,39 @@ func (client *Client) readResources(ctx context.Context) ([]*Edges, error) { //n
...
-func (client *Client) updateResource(ctx context.Context, resource *Resource) error {
+func (client *Client) updateResource(ctx context.Context, resource *Resource) (*Resource, error) {
 	variables := map[string]interface{}{
 		"id":              resource.ID,
 		"name":            resource.Name,
 		"address":         resource.Address,
 		"remoteNetworkId": resource.RemoteNetworkID,
 		"groupIds":        resource.GroupsIds,
 		"protocols":       resource.Protocols,
+		"first":           graphql.Int(readResourceQueryGroupsSize),
```
ekampf, line 284: same

ekampf (changes requested):
Looks good but require a small fix

twingate/client_resource.go
vmanilo: good point, thank you

vmanilo: updated

ekampf (approved):

Issue #203, opened 2022-10-14
Remove `readResourceQueryGroupsSize` limit


ekampf:
Currently, when reading a resource's groups we limit our query to `readResourceQueryGroupsSize` (50) which is essentially a single page in the resource's `groups` connection.

We should remove this limitation to support only 50 groups per resource and properly paginate the entire connection

PR #207, opened 2022-10-19
Fetch all resource groups

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/203

## Changes
- loading all groups for the resource

### DIFF
--- a/twingate/client_resource.go
+++ b/twingate/client_resource.go
@@ -23,6 +23,7 @@ type Resource struct {
 
 type gqlResource struct {
 	IDName
+	Groups  Groups `graphql:"groups(first: $groupsPageSize)"`
 	Address struct {
 		Type  graphql.String
 		Value graphql.String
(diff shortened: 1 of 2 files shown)


twingate/resource_resource_test.go
```diff
@@ -831,3 +832,82 @@ func createResource12(networkName, groupName1, groupName2, resourceName string)
...
+	}
+
+    %s
+
+	resource "twingate_resource" "test13" {
+	  name = "%s"
+	  address = "acc-test.com.13"
+	  remote_network_id = twingate_remote_network.test13.id
+	  group_ids = [%s]
+      protocols {
```
ekampf, line 880: Lets make indentation nice

twingate/client_resource.go
```diff
@@ -214,7 +222,12 @@ func (client *Client) createResource(ctx context.Context, resource *Resource) (*
 		return nil, NewAPIError(ErrGraphqlResultIsEmpty, "create", resourceResourceName)
 	}
 
-	return response.ResourceCreate.Entity.convertResource(), nil
+	resource, err := client.readAllResourceGroups(ctx, response.ResourceCreate.Entity)
```
ekampf, line 225: This is not necessary on `create` - you just created it you already know all its group ids...

ekampf (commented):
Something I dont understand here... maybe I don't understand the flow in TF...
You only updated the `create` and `update` operations but in those ops you already know the full list of group ids you want set...

Where we need pagination is during `read` when TF checks for drift?

twingate/client_resource.go
vmanilo: I assumed that on the backend side might be some additional steps like:
- assigning some additional system groups
- filtering out/not allow to assign some groups

if you think it's not the case, then we can really simply remove this call for fetching all pages

vmanilo:
> Something I dont understand here... maybe I don't understand the flow in TF... You only updated the `create` and `update` operations but in those ops you already know the full list of group ids you want set...
> 
> Where we need pagination is during `read` when TF checks for drift?

right, but we actually setting the state after create/update call successfully done, I mean that on the backend might be some business logic so result resource may have a different data compare to what we send in the request

twingate/resource_resource_test.go
vmanilo: fixed

ekampf (approved):

PR #210, opened 2022-10-24
Fix acc tests

ekampf:
Fixes acceptance test random failures

## Changes
Add a delay between the test's `config` phase and `check` phase. As check usually query replica DBs there's a chance of a small replication delay

### DIFF
--- a/twingate/resource_resource_test.go
+++ b/twingate/resource_resource_test.go
@@ -3,8 +3,10 @@ package twingate
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"regexp"
 	"testing"
+	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/resource"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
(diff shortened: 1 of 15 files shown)


twingate/resource_resource_test.go
```diff
@@ -24,6 +26,15 @@ const (
 	udpPorts       = "protocols.0.udp.0.ports.#"
 )
 
+func wait() resource.TestCheckFunc {
```
ekampf, line 29: @vmanilo whats the best place to put this function?  I have a feeling we'll need it on other ACC tests too...

vmanilo: maybe let's add to latest tests that were failing, I will do it

ekampf: @vmanilo no I mean what file should the `wait` function itself be on...

vmanilo: I think we can keep it in this file

vmanilo: it will be accessible in other acc tests, as it's the same package

liorr (approved):
LGTM

twingate/test_utils.go
```diff
@@ -0,0 +1,36 @@
...
+		// Sleep 500 ms
+		time.Sleep(WaitDuration)
+
+		return nil
+	}
+}
+
+func ComposeTestCheckFunc(fs ...resource.TestCheckFunc) resource.TestCheckFunc { //nolint:varnamelen
+	return func(state *terraform.State) error {
+		if err := WaitTestFunc()(state); err != nil {
```
vmanilo, line 24: not sure it's needed, as wait func never returns error

ekampf: Just for code completeness. If anyone ever goes and changes that func and doesn't look here

PR #211, opened 2022-10-24
Feature/fix acc tests

vmanilo:
Resolves acc issues

## Changes
- added custom test check func

### DIFF
--- a/twingate/resource_resource_test.go
+++ b/twingate/resource_resource_test.go
@@ -29,7 +29,7 @@ const (
 func wait() resource.TestCheckFunc {
 	return func(s *terraform.State) error {
 		// Sleep between 500ms-1s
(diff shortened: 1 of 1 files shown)


twingate/resource_resource_test.go
```diff
@@ -602,7 +616,8 @@ func TestAccTwingateResourceSetActiveStateOnUpdate(t *testing.T) {
 			},
 			{
 				Config: createResource10(remoteNetworkName, resourceName),
-				Check: resource.ComposeTestCheckFunc(
+				Check: ComposeTestCheckFunc(
+					wait(),
```
ekampf, line 620: This is a mistake... you already wait at the beginning in `ComposeTestCheckFunc`

vmanilo: agree

Issue #67, opened 2022-04-07
Add the ability to create and configure service accounts


bertekintw:
**Create a new `twingate_service_account` resource**

**New functionality**

Arguments

- (Required) `name` of type `string`
    - What the service account will be called in Twingate.

Notes

- We will not allow setting assigned resources here and instead require using the `twingate_resources_service_account_assign` and `twingate_resources_service_account_add` resources.

Example usage

```terraform
resource "twingate_service_account" "github_actions_prod" {
	name = "Github Actions PROD"
}
```

**Create a new `twingate_service_account_key` resource**

**New functionality**

Arguments

- (Required) `service_account_id` of type `string`
    - The `id` of the `twingate_service` resource that the key is assigned to.
- (Optional) `name` of type `string`
    - What the service key will be called in Twingate. If no name is assigned, Twingate will randomly generate a name.

Example usage

```terraform
resource "twingate_service_account" "github_actions_prod" {
  name = "Github Actions PROD"
}

resource "twingate_service__accountkey" "github_actions_prod" {
  service_id = twingate_service_account.github_actions_prod.id
}
```

badsector3:
Any updates on that? Do we have any ETA when can be expect the resource to be added to the provider?

alexmensch:
Hey @badsector3 we're working through the issues in Stage 1 and 2 milestones first, which we're thinking will take us into mid-July, and then we'll be tackling these issues in Stage 3 after that. We'll be making updates in this issue, so feel free to follow it.

ekampf:
@vmanilo lets start with the 2 resources:
- Resource `twingate_service_account`
  - field `name`
- Resource `twingate_service_account_key`
  - field `name`
  - field `service_account_id`
  - field `expiration_time` 


Afterwards we need to model the `service-account <> resource` relationship

vmanilo:
On the call we discussed changes for Resource `twingate_service_account_key`
* add `status` field
* remove `expiration_time`
* uses zero expiration value for now

alexmensch:
Hey @vmanilo I updated the issue description to reflect the changes we'd like to make here. Just some minor changes (eg. the name of the resource), otherwise should be in line with the PRs you're working on already.

vmanilo:
got it, will update resource and field names

PR #219, opened 2022-11-16
Resource: service

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/67

## Changes
- added service-account resource


twingate/internal/provider/resource/service-account.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/Twingate/terraform-provider-twingate/twingate/internal/client"
+	"github.com/Twingate/terraform-provider-twingate/twingate/internal/model"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func ServiceAccount() *schema.Resource {
+	return &schema.Resource{
+		Description:   "Services offer a way to provide programmatic, centrally-controlled, and consistent access controls. For more information, see Twingate's [documentation](https://www.twingate.com/docs/services).",
```
vmanilo, line 16: @alexmensch please check description

docs/resources/service_account.md
```diff
@@ -0,0 +1,37 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "twingate_service_account Resource - terraform-provider-twingate"
+subcategory: ""
+description: |-
+  Services offer a way to provide programmatic, centrally-controlled, and consistent access controls. For more information, see Twingate's documentation https://www.twingate.com/docs/services.
```
alexmensch, line 6: I know this is taken from the public docs, but let's do this everywhere: s/Services/Service Accounts/g

In this description, let's remove the sentence "For more information [...]"

docs/resources/service_account.md
```diff
@@ -0,0 +1,37 @@
...
+---
+
+# twingate_service_account (Resource)
+
+Services offer a way to provide programmatic, centrally-controlled, and consistent access controls. For more information, see Twingate's [documentation](https://www.twingate.com/docs/services).
```
alexmensch, line 11: s/Services/Service Accounts/

docs/resources/service_account.md
```diff
@@ -0,0 +1,37 @@
...
+  name = "Github Actions PROD"
+}
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `name` (String) The name of the service account in Twingate
```
alexmensch, line 31: s/service account/Service Account/

docs/resources/service_account.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Read-Only
+
+- `id` (String) Autogenerated ID of the service account
```
alexmensch, line 35: s/service account/Service Account/

alexmensch (changes requested):
Some simple changes (I'll make the same changes in the public documentation)

docs/resources/service_account.md
vmanilo: @alexmensch I've updated doc strings

docs/resources/service_account.md
vmanilo: updated

docs/resources/service_account.md
vmanilo: updated

docs/resources/service_account.md
vmanilo: updated

twingate/internal/client/pagination_test.go
```diff
@@ -0,0 +1,84 @@
+package client
```
ekampf, line 1: Why is this here and not in `twingate/internal/test/client` ?

twingate/internal/client/converter.go
```diff
@@ -230,3 +230,31 @@ func (r ResourceNode) ToModel() *model.Resource {
 func (q readResourcesByNameQuery) ToModel() []*model.Resource {
 	return q.Resources.ToModel()
 }
+
```
ekampf, line 233: Its something I missed in the restructure review but Im not sure I understand the purpose of `converter.go` - we define `createServiceAccountQuery` in `service-account.go` why cant its methods be there?  
(same for all the other structs that their methods are here)

twingate/internal/client/pagination_test.go
vmanilo: here I'm testing private func `fetchPages` and it's not accessible from other packages

twingate/internal/client/converter.go
vmanilo: here in `converter.go` I've collected simple utilities functions which converts queries objects to models

I think in `service-account.go` would be better to have only code related to API request, and this object conversions like lower layer details

I like the idea with separate package for `queries`, I think this can help organise code better :)

ekampf: ok so lets merge this and refactor to `queries` on a separate PR

Issue #231, opened 2022-12-09
Add location property to RemoteNetwork


ekampf:

vmanilo:
@ekampf @alexmensch please specify where is new property should be added to datasource or resource or both

PR #232, opened 2022-12-10
Feature: added remote network `location` property

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/231

## Changes
- added `location` property to remote-network resource and datasource

### DIFF
--- a/twingate/internal/model/remote-network.go
+++ b/twingate/internal/model/remote-network.go
@@ -8,8 +8,7 @@ const (
 	LocationOther       = "OTHER"
 )
 
-// nolint
-var Locations = []string{LocationAWS, LocationAzure, LocationGoogleCloud, LocationOnPremise, LocationOther}
+var Locations = []string{LocationAWS, LocationAzure, LocationGoogleCloud, LocationOnPremise, LocationOther} //nolint
 
 type RemoteNetwork struct {
 	ID       string
--- a/twingate/internal/model/remote-network.go
+++ b/twingate/internal/model/remote-network.go
@@ -8,7 +8,7 @@ const (
 	LocationOther       = "OTHER"
 )
 
-//nolint
+// nolint
 var Locations = []string{LocationAWS, LocationAzure, LocationGoogleCloud, LocationOnPremise, LocationOther}
 
 type RemoteNetwork struct {
--- a/twingate/internal/model/remote-network.go
+++ b/twingate/internal/model/remote-network.go
@@ -1,8 +1,20 @@
 package model
 
+const (
+	LocationAWS         = "AWS"
+	LocationAzure       = "AZURE"
+	LocationGoogleCloud = "GOOGLE_CLOUD"
+	LocationOnPremise   = "ON_PREMISE"
+	LocationOther       = "OTHER"
+)
+
+//nolint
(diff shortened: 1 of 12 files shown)


twingate/internal/provider/datasource/remote-network.go
```diff
@@ -44,6 +44,11 @@ func RemoteNetwork() *schema.Resource {
 				Description:  "The name of the Remote Network",
 				ExactlyOneOf: []string{"id"},
 			},
+			"location": {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "The location of the Remote Network",
```
vmanilo, line 50: @alexmensch please check description

twingate/internal/provider/resource/remote-network.go
```diff
@@ -30,6 +31,13 @@ func RemoteNetwork() *schema.Resource {
 				Required:    true,
 				Description: "The name of the Remote Network",
 			},
+			"location": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringInSlice(model.Locations, false),
+				Description:  "The location of the Remote Network",
```
vmanilo, line 38: @alexmensch please check description

twingate/internal/model/remote-network.go
```diff
@@ -1,8 +1,19 @@
...
 
+const (
+	LocationAWS         = "AWS"
+	LocationAzure       = "AZURE"
+	LocationGoogleCloud = "GOOGLE_CLOUD"
+	LocationOnPremise   = "ON_PREMISE"
+	LocationOther       = "OTHER"
+)
+
+var Locations = []string{LocationAWS, LocationAzure, LocationGoogleCloud, LocationOnPremise, LocationOther} //nolint
```
ekampf, line 11: What happens if we add a new Enum value on server side?   will it break TF?

vmanilo: new enum value won’t break it
* datasource accepts any value as they fetched from backend side (so we don’t need to validate it)
* on resource side - user should provide this value and terraform validates according to our list of possible locations, so when new enum added on server side - terraform just won’t allow to use this new value until we update list of allowed location options

twingate/internal/provider/datasource/remote-network.go
alexmensch: We should list out the possible options here. So let's change it to: `The location of the Remote Network. Must be one of the following: AWS, AZURE, GOOGLE_CLOUD, ON_PREMISE, OTHER`

twingate/internal/provider/resource/remote-network.go
alexmensch: We should list out the possible options here. So let's change it to: The location of the Remote Network. Must be one of the following: AWS, AZURE, GOOGLE_CLOUD, ON_PREMISE, OTHER

twingate/internal/provider/datasource/remote-network.go
alexmensch: The only downside to this is that the docs aren't automatically up to date with our API, but I don't think these are going to change very often.

alexmensch (changes requested):
Just need to provide a list of what the available options are somewhere.

alexmensch (approved):
Nice, I like your solution to list them out programmatically :)

ekampf (approved):

Issue #228, opened 2022-12-06
Allow lazy client initialization


alexmensch:
Do not initialize the client process if the API token has not been configured.

PR #233, opened 2022-12-10
Feature: client lazy initialization

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/228

## Changes
- changed client initialization, so api token can be skipped


twingate/internal/client/client.go
```diff
@@ -61,9 +92,19 @@ func newServerURL(network, url string) serverURL {
...
+		return false, err
+	}
+
+	return retryablehttp.DefaultRetryPolicy(ctx, resp, err) //nolint
+}
+
 func NewClient(url string, apiToken string, network string, httpTimeout time.Duration, httpRetryMax int, version string) *Client {
 	sURL := newServerURL(network, url)
 	retryableClient := retryablehttp.NewClient()
+	retryableClient.CheckRetry = customRetryPolicy
```
ekampf, line 107: If I understand this code correctly what will happen if I run TF without a Twingate API key is that you'll still make a ton of HTTP requests they will just fail and we ignore that failure?
That's not a good solution...

ekampf (approved):

Issue #242, opened 2022-12-16
Deprecate the `group_ids` argument in the `twingate_resource` resource


alexmensch:
Issue #241 creates a new resource for access assignment, which conflicts with the existing group assignment functionality in the `twingate_resource` resource. We'll deprecate this functionality and remove it entirely at a later date in #225.

alexmensch:
Re-opening this to align with our release planning.

vmanilo:
@alexmensch this changes already in the main branch 

https://github.com/Twingate/terraform-provider-twingate/blob/main/twingate/internal/provider/resource/resource.go#L120

alexmensch:
Ah, thanks! Closing

PR #246, opened 2022-12-20
Deprecated group_ids attribute

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/242

## Changes
- deprecated attribute `group_ids`


ekampf (approved):

twingate/internal/provider/resource/resource.go
```diff
@@ -66,6 +66,7 @@ func Resource() *schema.Resource { //nolint:funlen
 				Elem:        &schema.Schema{Type: schema.TypeString},
 				Optional:    true,
 				Description: "List of Group IDs that have permission to access the Resource, cannot be generated by Terraform and must be retrieved from the Twingate Admin Console or API",
+				Deprecated:  "Configure `access` block instead. This attribute will be removed in the next version of the provider.",
```
alexmensch, line 69: Let's change this to: "The group_ids argument is now deprecated, and the new access block argument should be used instead. The group_ids argument will be removed in a future version of the provider."

alexmensch (changes requested):

twingate/internal/provider/resource/resource.go
vmanilo: updated

alexmensch (approved):

tjstansell:
Hm... odd that you'd merge this before an `access {}` block even exists.  I'm used to seeing new features added and tested for a few releases and _then_ the old mechanism deprecated. :)

rebmcr:
> Hm... odd that you'd merge this before an `access {}` block even exists. I'm used to seeing new features added and tested for a few releases and _then_ the old mechanism deprecated. :)

Indeed, looks like the `access{}` block might fulfill a feature request we made, so hopefully this is just out-of-sequence comms!

alexmensch:
This PR just marks the argument as deprecated, we won't actually remove support until later on.

The next release will include this plus PR #249 so they will be released together. Merge != release!

PR #247, opened 2022-12-20
Release as draft first

### LINKED ISSUE #244: automate creation of CHANGELOG.md and/or releases
The last 2 releases for this provider haven't included great changelogs.  v0.2.2 only mentioned a doc update, but actually included support for locations in remote networks as well (which was broken).  v0.2.3 was just released and mentions the fix for location, but the release doesn't include links to the PR.  Neither of these follow the release formatting done previously.  Perhaps some automation would be better suited for release management and ideally a running CHANGELOG.md could be built so that it's easy to see the history of changes instead of having to browse the github releases?  I would suggest looking at adopting Hashicorp's go-changelog, like AWS has used:

https://github.com/hashicorp/terraform-provider-aws/blob/main/.github/workflows/generate_changelog.yml

Any other solution would be fine if it provides more complete and consistent changelogs to see what has changed between releases.

ekampf:
Resolves #244 

## Changes
- Make new releases as drafts so we publish them after Changelog is written

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -42,7 +42,6 @@ signs:
       - "--detach-sign"
       - "${artifact}"
 release:
-# If you want to manually examine the release before its live, uncomment this line:
-# draft: true
+  draft: true
 changelog:
   skip: true


romankor (approved):
GTM

Issue #245, opened 2022-12-20
Create a new `access` block in `twingate_resource` for group and service account assignment


alexmensch:
**Modified functionality**

Arguments

- (Optional) `access` block to define which groups and service accounts have access to the Twingate resource.
    - (Optional) `group_ids` of type `list(string)`
        - List of resource `id`s obtained from the `twingate_group` or `twingate_groups` data sources.
    - (Optional) `service_account_ids` of type `list(string)`
        - List of resource `id`s from service account resources
- (Optional) `authoritative` of type `boolean`
    - Default: `true`

Behavior

- We are deprecating the `group_ids` argument outside of the new `access` block in #242, which means that we must generate a blocking error if both `group_ids` and `access` are specified. Only one is allowed as they are mutually exclusive options after this change.
- Although both `group_ids` and `service_account_ids` are individually optional, at least one must be specified for the `access` block to be valid.
- The `authoritative` argument determines whether changes made outside of Terraform are overridden by Terraform. By default, Terraform is the authoritative source of true. When set to `false`, changes made outside Terraform are ignored.
- Only a single access block is allowed.

Example usage

```terraform
data "twingate_groups" "devops" {
  name = "DevOps"
}

resource "twingate_service_account" "github_actions_prod" {
  name = "Github Actions PROD"
}

resource "twingate_remote_network" "aws" {
  name = "AWS"
}

resource "twingate_resource" "internal_resources" {
  name = "All internal resources"
  address = "*.internal.int"
  remote_network_id = twingate_remote_network.aws.id
  protocols {
    allow_icmp = true
    tcp  {
      policy = "RESTRICTED"
      ports = ["80", "82-83"]
    }
    udp {
      policy = "ALLOW_ALL"
    }
  }
  access {
    service_account_ids = toset([twingate_service_account.github_actions_prod.id])
    group_ids = toset([data.twingate_groups.devops.id])
  }
  authoritative = false
}
```

ekampf:
Yes lets enforce a single one for now...

alexmensch:
Done

alexmensch:
Making one more change to the spec here. Discussed with @ekampf and @vmanilo today.

1. We're pulling the `authoritative` argument outside of the individual `access` block.
- The rationale for this is that although there is no use case for having multiple `access` blocks today, we are currently working on making changes to our core policy framework that will allow assigning policies to individual access assignments. Once this is available in the core product we will:
[...]

PR #249, opened 2022-12-22
Feature: Resource access block

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/245, https://github.com/Twingate/terraform-provider-twingate/issues/242

## Changes
- added new property `access` block with authoritative flag


alexmensch (approved):

tjstansell:
Isn't group_ids authoritative today? Shouldn't access blocks have the same behavior?

alexmensch:
> Isn't group_ids authoritative today? Shouldn't access blocks have the same behavior?

We don't want them authoritative because we want to balance allowance for manual updates made via the admin console. We may add an authoritative flag later on. If you exclusively make changes to your Twingate instance via TF it's effectively authoritative.

tjstansell:
That seems backwards. If you are adding resources via terraform I would expect it to have authoritative control over who should be allowed to access it. It's fine to have an option or method to disable that, but it shouldn't be the default, IMO. and the point of having it in terraform is to ensure the config is as defined in code. We WANT it to revert anything someone changes in the UI.

vmanilo:
@ekampf @alexmensch please take a look, I've added new tests and changed logic of handling groups and service accounts for the resource access block.

tjstansell:
If you already have the logic to toggle between authoritative and non-authoritative, then please put that behind a new resource argument to lets folks opt into that. You shouldn't change the default behavior to be non-authoritative. And without a way to force authoritative control, we will have to pin our provider to an old version (the current one) until it does exist.

alexmensch:
> That seems backwards. If you are adding resources via terraform I would expect it to have authoritative control over who should be allowed to access it. It's fine to have an option or method to disable that, but it shouldn't be the default, IMO. and the point of having it in terraform is to ensure the config is as defined in code. We WANT it to revert anything someone changes in the UI.

Understood. We discussed today and will be adding in a flag before this ends up being released, with the default being the existing authoritative behavior. That will retain proper backwards compatibility.

docs/resources/resource.md
```diff
@@ -55,13 +64,24 @@ resource "twingate_resource" "resource" {
...
 
 - `id` (String) Autogenerated ID of the Resource, encoded in base64
 
+<a id="nestedblock--access"></a>
+### Nested Schema for `access`
+
+Optional:
+
+- `group_ids` (Set of String) List of Group IDs that have permission to access the Resource.
+- `non_authoritative` (Boolean) Determines authoritative behaviour for handling the Resource's groups or service accounts.
```
tjstansell, line 81: I'm curious how this would work being assigned to an access block instead of to the resource itself.  It seems this should be set at the resource level, because eventually this will support multiple access blocks and it doesn't make sense for some to be "authoritative" and others to not be.  But maybe I'm mis-understanding some use case?

vmanilo: I think multiple "authoritative" access blocks will make conflicts with each other, multiple non-authoritative blocks can be merged into one.

Currently put "authoritative" flag inside access block, as it's logically relates more to access settings than to resource itself. 

I guess multiple access blocks would make more sense with policies.

twingate/internal/provider/resource/resource.go
```diff
@@ -37,6 +39,60 @@ func Resource() *schema.Resource { //nolint:funlen
...
+				MinItems:     1,
+				Optional:     true,
+				AtLeastOneOf: []string{attr.Path(attr.Access, attr.GroupIDs)},
+				Description:  "List of Service Account IDs that have permission to access the Resource.",
+			},
+			attr.NonAuthoritative: {
+				Type:        schema.TypeBool,
+				Optional:    true,
+				Default:     false,
+				Description: "Determines authoritative behaviour for handling the Resource's groups or service accounts.",
```
vmanilo, line 91: @alexmensch please check docs string

docs/resources/resource.md
tjstansell: I see that @alexmensch had originally specified it as part of the `access {}` block in the original issue.  I'm curious if that was just a mistake or if it was on purpose.  I don't see how it would work once you support multiple access blocks.

ekampf, line 81: @vmanilo why is the property called `non_authoritative` ? marking it false is double negative - not not authoritative?  it should just be straightforward `authoritative`

vmanilo: yes, we want to have `authoritative` mode by default, means it should be `true` by default even when user not set it manually, in that case terraform will show drift because in the state will be added  authoritative=true

alexmensch: Hey @tjstansell we've clarified this behavior here: https://github.com/Twingate/terraform-provider-twingate/issues/245#issuecomment-1386003243

Bit of internal back and forth missed communication in the rush before the holiday break. The whole team is back now, so we're back at it! :)

docs/resources/resource.md
```diff
@@ -55,7 +64,9 @@ resource "twingate_resource" "resource" {
 
 ### Optional
 
-- `group_ids` (Set of String) List of Group IDs that have permission to access the Resource, cannot be generated by Terraform and must be retrieved from the Twingate Admin Console or API
+- `access` (Block List, Max: 1) Restrict access to certain groups or service accounts (see [below for nested schema](#nestedblock--access))
+- `authoritative` (Boolean) Determines authoritative behaviour for handling the Resource's groups or service accounts.
```
alexmensch, line 68: --> "Determines whether assignments in the `access` block will override any existing assignments. Default is `true`. If set to `false`, assignments made outside of Terraform will be ignored."

alexmensch, line 68: This argument also needs to be named `is_authoritative` in line with other Boolean arguments that existing in our provider.

docs/resources/resource.md
```diff
@@ -64,6 +75,15 @@ resource "twingate_resource" "resource" {
 
 - `id` (String) Autogenerated ID of the Resource, encoded in base64
 
+<a id="nestedblock--access"></a>
+### Nested Schema for `access`
+
+Optional:
+
+- `group_ids` (Set of String) List of Group IDs that have permission to access the Resource.
```
alexmensch, line 83: --> "List of Group IDs that must have permission to access the Resource."

docs/resources/resource.md
```diff
@@ -64,6 +75,15 @@ resource "twingate_resource" "resource" {
...
+- `service_account_ids` (Set of String) List of Service Account IDs that have permission to access the Resource.
```
alexmensch, line 84: --> "List of Service Account IDs that must have permission to access the Resource."

alexmensch (changes requested):

vmanilo:
updated docs according to requested changes

alexmensch (approved):

twingate/internal/utils/utils.go
```diff
@@ -22,3 +23,32 @@ func Filter[T any](items []T, ok func(item T) bool) []T {
...
+
+// Contains - checks if element exists in the slice.
+func Contains[T comparable](items []T, element T) bool {
+	lookup := MakeLookupMap[T](items)
+
+	return lookup[element]
+}
+
+// MapKeys - collects map keys to slice.
+func MapKeys[T comparable](lookup map[T]bool) []T {
```
ekampf, line 35: There's something fishy about these 2 methods - `MapKeys`, `MakeLookupMap` - converting a list of values to a map of `value -> true` and back.
Doesn't Golang has a `set` ?

ekampf: Bah thats the Golang way it seems...

twingate/internal/client/resource.go
```diff
@@ -320,3 +328,113 @@ func (client *Client) readResourcesByNameAfter(ctx context.Context, variables ma
...
+	return nil
+}
+
+func (client *Client) ReadResourceServiceAccounts(ctx context.Context, resourceID string) ([]string, error) {
+	serviceAccounts, err := client.ReadServiceAccounts(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	serviceAccountIDs := make(map[string]bool)
```
ekampf, line 417: Why is it a map you need to convert back to list in line 425 and not just a list you append to?

ekampf: Are you trying to de-dupe duplicate service account ids here? why?

twingate/internal/utils/utils.go
ekampf, line 35: Maybe we should just include a `Set` implementation dependency?   would make the code nicer?

twingate/internal/utils/utils.go
```diff
@@ -22,3 +23,32 @@ func Filter[T any](items []T, ok func(item T) bool) []T {
 
 	return result
 }
+
+// Contains - checks if element exists in the slice.
+func Contains[T comparable](items []T, element T) bool {
+	lookup := MakeLookupMap[T](items)
```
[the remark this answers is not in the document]
vmanilo: yes, agree it will be pretty the same, the idea was just to re-use existing code

but let's do as you propose, as we don't need to allocate extra memory for this

twingate/internal/utils/utils.go
vmanilo: yeah, golang doesn't has `set`

docs/resources/resource.md
```diff
@@ -64,6 +75,15 @@ resource "twingate_resource" "resource" {
 
 - `id` (String) Autogenerated ID of the Resource, encoded in base64
 
+<a id="nestedblock--access"></a>
+### Nested Schema for `access`
+
+Optional:
+
+- `group_ids` (Set of String) List of Group IDs that must have permission to access the Resource.
```
alexmensch, line 83: Hey @vmanilo sorry I made a mistake here, should be "will" instead of "must". Full sentence: "List of Group IDs that will have permission to access the Resource."

twingate/internal/client/resource.go
vmanilo: yes, I think here we don't need to do de-duplication

ekampf (approved):
Fix and merge

docs/resources/resource.md
vmanilo: @alexmensch what about `service_account_ids`, do I need to fix it as well?

currently it looks like this: `List of Service Account IDs that must have permission to access the Resource.`

alexmensch: Ah, yes please @vmanilo! Yeah, I'm not sure what I was thinking last night, but it should be "will" instead of "must"

Issue #259, opened 2023-01-17
Allow user assignment to `twingate_group` resource


alexmensch:
**Modified functionality**

Arguments

- (Optional) `user_ids` specifies the set of users that must be assigned to this group. User IDs can be obtained from the `twingate_user` and `twingate_users` data sources.
- (Optional) `is_authoritative` specifies whether the set of assigned users must override any existing assignment(s) or changes made outside of the Terraform configuration.
  - Defaults to `true`.

Behavior

- The `is_authoritative` flag follows the same behavior pattern as in issue #245.

Example usage

```terraform
data "twingate_users" "all" { }

locals {
   users_admins = [ for i in data.twingate_users.all.users : i if i.is_admin == true ]
}

resource "twingate_group" "admins" {
  name = "Admins"
  user_ids = local.users_admins
  is_authoritative = true
}
```

PR #266, opened 2023-01-29
Feature: allow user assignment to twingate_group

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/259

## Changes
- added optional properties to twingate_group: authoritative flag and users list


flatheadmill:
🙏🏻.

docs/resources/group.md
```diff
@@ -30,6 +30,11 @@ resource "twingate_group" "aws" {
 
 - `name` (String) The name of the group
 
+### Optional
+
+- `authoritative` (Boolean) Determines authoritative behaviour for handling the Group's users.
```
alexmensch, line 35: --> "Determines whether User assignments to this Group will override any existing assignments. Default is `true`. If set to `false`, assignments made outside of Terraform will be ignored."

alexmensch, line 35: This argument should be named `is_authoritative` since it's a Boolean

alexmensch (changes requested):

vmanilo:
updated docs

alexmensch (approved):

docs/resources/resource.md
```diff
@@ -64,6 +75,15 @@ resource "twingate_resource" "resource" {
 
 - `id` (String) Autogenerated ID of the Resource, encoded in base64
 
+<a id="nestedblock--access"></a>
+### Nested Schema for `access`
+
+Optional:
+
+- `group_ids` (Set of String) List of Group IDs that must have permission to access the Resource.
```
alexmensch, line 83: Hey @vmanilo sorry I made a mistake here, should be "will" instead of "must". Full sentence: "List of Group IDs that will have permission to access the Resource."

vmanilo: fixed

ekampf (changes requested):
Need to check group type is `MANUAL` before trying to make any change to users membership. If its not - error

ekampf:
@alexmensch if I import a group thats not `MANUAL` to TF and then try to change its user membership what should the error text be?
Something like "You can only change user membership on MANUAL groups"

twingate/internal/provider/resource/helper.go
```diff
@@ -24,3 +24,37 @@ func convertIDs(data interface{}) []string {
 		},
 	)
 }
+
+// setIntersection - for given two sets A and B,
+// A ∩ B (read as A intersection B) is the set of common elements that belong to set A and B.
+// If A = {1, 2, 3, 4} and B = {3, 4, 5, 7}, then the intersection of A and B is given by A ∩ B = {3, 4}.
+func setIntersection(a, b []string) []string {
```
ekampf, line 31: need to move its tests from `resource_test.go` to a new `helpers_test.go`

twingate/internal/provider/resource/helper.go
```diff
@@ -24,3 +24,37 @@ func convertIDs(data interface{}) []string {
...
+		}
+	}
+
+	return result
+}
+
+// setDifference - difference between sets implies subtracting the elements from a set.
+// The difference between sets A and set B denoted as A − B.
+// If A = {1, 2, 3, 4} and B = {3, 4, 5, 7}, then the difference between sets A and B is given by A - B = {1, 2}.
+func setDifference(a, b []string) []string {
```
ekampf, line 48: Lets add test

ekampf (approved):
Add small fixes and we can merge

twingate/internal/provider/resource/group.go
```diff
@@ -3,14 +3,18 @@ package resource
...
 	"log"
 
+	"github.com/Twingate/terraform-provider-twingate/twingate/internal/attr"
 	"github.com/Twingate/terraform-provider-twingate/twingate/internal/client"
 	"github.com/Twingate/terraform-provider-twingate/twingate/internal/model"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 )
 
+var ErrAllowedToChangeOnlyManualGroups = fmt.Errorf("allowed to change only %s groups", model.GroupTypeManual)
```
vmanilo, line 16: @alexmensch please check this message

twingate/internal/provider/resource/helper.go
vmanilo: moved test to helpers_test

twingate/internal/provider/resource/helper.go
vmanilo: added test

twingate/internal/provider/resource/group.go
alexmensch: Oh, I see. Okay! Let's do this:

"Only groups of type %s may be modified"

The problem is that I think most people don't know what a "MANUAL" group is for this error. Could we also show what the type of the group is where the attempt was made?

So something like this: ("Only groups of type %s may be modified. Group %s is a %s type group.", model.GroupTypeManual, <groupName>, <groupType>)

vmanilo: sure

vmanilo: updated error message

alexmensch (approved):

ekampf (approved):

flatheadmill:
🎉🍾🥂

PR #286, opened 2023-02-26
[TBD] Test: not allow update system group

vmanilo:
## Changes
- added acc test to check error message on update system group


go.mod
```diff
@@ -16,6 +16,8 @@ require (
 	gotest.tools/gotestsum v1.9.0
 )
 
+replace github.com/hashicorp/terraform-plugin-sdk/v2 v2.25.2 => github.com/vmanilo/terraform-plugin-sdk/v2 v2.25.2
```
ekampf, line 19: I dont want to use a fork of the SDK so we need to find away around this

ekampf (changes requested):
lets try and figure out a way that doesn't require terraform-sdk fork

go.mod
vmanilo: I've added PR to the terraform-plugin-sdk https://github.com/hashicorp/terraform-plugin-sdk/pull/1159

let's see if it be excepted and released

vmanilo:
@ekampf FYI terraform guys extracted testing framework from sdk repo

https://discuss.hashicorp.com/t/announcing-the-terraform-plugin-testing-module/49914

they suggest to switch to new repo https://github.com/hashicorp/terraform-plugin-testing as it will be continue developing, and probably later there will be way to prevent destroy resource after tests.

so not sure, should we close this PR for now or just leave it as draft?

vmanilo:
let's wait for PR with new option `RemoveState` https://github.com/hashicorp/terraform-plugin-testing/pull/118

vmanilo:
PR to `terraform-plugin-testing` was closed https://github.com/hashicorp/terraform-plugin-testing/pull/118#issuecomment-1658843444

probably later this year terraform will have native support to unmanage resources, then we can get back to this test

Issue #291, opened 2023-03-02
Allow controlling status notifications for the `twingate_connector` resource


alexmensch:
The Twingate API exposes parameters that allow controlling whether [status notifications](https://www.twingate.com/docs/connectors/#status-availability-emails) are sent for Connectors:

- `connectorCreate(hasStatusNotificationsEnabled = true)`
- `connectorUpdate(hasStatusNotificationsEnabled = null)`

This issue proposes adding this capability to the `twingate_connector` resource and data source.

**New optional argument: `status_updates_enabled`**

`twingate_connector` resource:
- This argument must be added to the `twingate_connector` resource.
- If the argument is not specified, the default value must match the API behavior (which is `true` in this case, but in principle could be changed at the API level in the future).

`twingate_connector` data source:
- This value must be added to the `twingate_connector` data source and reflect the currently set value for the connector.

Migration behavior:
- The above changes should not introduce any plan changes for existing Terraform configurations.

PR #293, opened 2023-03-04
Added status_updates_enabled attribute to twingate_connector 

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/291

## Changes
- added new optional attribute `status_updates_enabled`


twingate/internal/client/connector.go
```diff
@@ -10,57 +10,92 @@ import (
...
 
-func (client *Client) CreateConnector(ctx context.Context, remoteNetworkID, connectorName string) (*model.Connector, error) {
-	if remoteNetworkID == "" {
+func (client *Client) CreateConnector(ctx context.Context, input *model.Connector) (*model.Connector, error) {
+	if input == nil || input.NetworkID == "" {
 		return nil, NewAPIError(ErrGraphqlNetworkIDIsEmpty, "create", connectorResourceName)
 	}
 
+	if input.StatusUpdatesEnabled != nil {
+		return client.createConnectorWithNotificationStatus(ctx, input)
```
ekampf, line 19: This doesn't seem like the right solution...  what if we had more optional variables?   we'll need to create a query struct and function for every possible permutation?

twingate/internal/client/query/connector-create.go
```diff
@@ -4,6 +4,10 @@ type CreateConnector struct {
 	ConnectorEntityResponse `graphql:"connectorCreate(remoteNetworkId: $remoteNetworkId, name: $connectorName)"`
 }
 
+type CreateConnectorWithNotificationStatus struct {
+	ConnectorEntityResponse `graphql:"connectorCreate(remoteNetworkId: $remoteNetworkId, name: $connectorName, hasStatusNotificationsEnabled: $hasStatusNotificationsEnabled)"`
```
ekampf, line 8: does `go-graphql` support defining [defaults](https://graphql.org/learn/queries/#default-variables) here?
```
graphql:"connectorCreate(remoteNetworkId: $remoteNetworkId, name: $connectorName, hasStatusNotificationsEnabled: $hasStatusNotificationsEnabled = true)"
```

vmanilo: no, it's not supported

twingate/internal/client/connector.go
vmanilo: we can set default value like this, 
but the downside that if you change default value on the backend side, we would need to update client as well

```
        statusUpdatesEnabled := true	// default value
	if input.StatusUpdatesEnabled != nil {
		statusUpdatesEnabled = *input.StatusUpdatesEnabled
	}

	variables := newVars(
		gqlID(input.NetworkID, "remoteNetworkId"),
		gqlNullable(input.Name, "connectorName"),
		gqlVar(statusUpdatesEnabled, "hasStatusNotificationsEnabled"),
	)
```

vmanilo: in the issue there is one point listed

```
If the argument is not specified, the default value must match the API behavior 

(which is true in this case, but in principle could be changed at the API level in the future).
```

vmanilo: @ekampf it's fixed now, thank you for the backend support :)

ekampf (approved):

PR #301, opened 2023-03-06
Fix acc tests terraform versions (Remove ACC tests on EOL TF versions)

### LINKED ISSUE #300: Remove EOL Terraform versions from ACC tests

ekampf:
Resolves #300 

## Changes
Based on https://endoflife.date/terraform versions 1.0 and 1.1 are now dead so we should be testing on 1.2, 1.3 and latest
Screenshot 2023-03-06 at 1 42 49 PM


romankor (approved):
Gtm

PR #304, opened 2023-03-07
Fix smoketests to not test on deprecated SDK versions

### LINKED ISSUE #300: Remove EOL Terraform versions from ACC tests

ekampf:
Resolves #300


liorr (approved):
LGTM

Issue #321, opened 2023-04-20
Add support for new `twingate_user` resource


alexmensch:
**New functionality**

### Arguments

**Only applied on user creation**
- `email` of type `string`
- (Optional) `send_invite` of type `boolean`
    - API default: `true`

**Updated on every plan application**
- (Optional) `first_name` of type `string`
    - API default: `null`
    - If not specified, the existing product behavior is to populate this information on first sign in.
- (Optional) `last_name` of type `string`
    - API default: `null`
    - If not specified, the existing product behavior is to populate this information on first sign in.
- (Optional) `role` which is an enum of the available roles exposed in the public admin API (currently ADMIN, DEVOPS, SUPPORT, MEMBER)
    - API default: `MEMBER`
- (Optional) `is_active` of type `boolean`
    - API default: `null`

### Behavior notes

- We'll add documentation to clarify the behavior to be expected due to the overlap with IdP-managed users, but otherwise the API will already return the various error states.

### Example usage

```terraform
resource "twingate_user" "external" {
  first_name = "Jane"
  last_name = "Doe"
  email = "jane_doe@acme.co"
  role = ADMIN
  send_invite = false
  is_active = true
}
```

PR #323, opened 2023-04-27
Resource: twingate_user

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/321

## Changes
- added new resource twingate_user
- added unit and acc tests


twingate/internal/provider/resource/user.go
```diff
@@ -0,0 +1,231 @@
...
+			attr.Type: {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: fmt.Sprintf("Indicates the User's type. Either %s.", utils.DocList(model.UserTypes)),
+			},
+			attr.ID: {
+				Type:        schema.TypeString,
+				Computed:    true,
+				Description: "Autogenerated ID of the User, encoded in base64.",
+			},
```
vmanilo, line 76: @alexmensch please take a look at descriptions

alexmensch (approved):
Made copy edits, good to go!

twingate/internal/attr/user.go
```diff
@@ -1,10 +1,12 @@
...
-	FirstName = "first_name"
-	LastName  = "last_name"
-	Email     = "email"
-	IsAdmin   = "is_admin"
-	Role      = "role"
-	Users     = "users"
+	FirstName  = "first_name"
+	LastName   = "last_name"
+	Email      = "email"
+	IsAdmin    = "is_admin"
```
ekampf, line 7: Why do we need this if we have `role` ?

vmanilo: not sure, maybe for easy filtering when we use datasource
in the first implementation that we have for user datasource from 16/08/2022 we already have this two fields

vmanilo: found this commit https://github.com/Twingate/terraform-provider-twingate/commit/81807422cac74a92cb61107f83e388bfb8e5b6a6

ekampf (approved):

PR #328, opened 2023-05-09
[Need to slice it down for smaller PRs] Upgrade provider with new TF plugin framework

vmanilo:
Resolves upgrading provider to new TF plugin framework

## Changes
- update resources and datasource with new TF plugin framework


docs/data-sources/service_accounts.md
```diff
@@ -27,10 +27,10 @@ data "twingate_service_accounts" "foo" {
 
 ### Read-Only
 
-- `id` (String) The ID of this resource.
-- `service_accounts` (Block List) List of Service Accounts (see [below for nested schema](#nestedblock--service_accounts))
+- `id` (String) The ID of the Service Accounts datasource.
```
alexmensch, line 30: Let's change this string to this: "The ID of the data source result"

You can use this on all data sources.

docs/data-sources/service_accounts.md
```diff
@@ -27,10 +27,10 @@ data "twingate_service_accounts" "foo" {
...
+- `service_accounts` (Attributes List) List of Service Accounts (see [below for nested schema](#nestedatt--service_accounts))
 
-<a id="nestedblock--service_accounts"></a>
+<a id="nestedatt--service_accounts"></a>
 ### Nested Schema for `service_accounts`
 
 Read-Only:
```
alexmensch, line 36: For the `id` value we should leave this as "The ID of the Service Account resource" since this is an output, not an input. This is the Terraform ID, correct?

docs/resources/group.md
```diff
@@ -38,7 +38,7 @@ resource "twingate_group" "aws" {
 
 ### Read-Only
 
-- `id` (String) Autogenerated ID of the Resource, encoded in base64
+- `id` (String) The ID of the Group
```
alexmensch, line 41: We should change this to "The ID of the Group resource" since this is an output. This is the Terraform ID, correct?

docs/resources/service_account_key.md
```diff
@@ -42,6 +42,7 @@ resource "twingate_service_account_key" "github_key" {
 ### Read-Only
 
 - `id` (String) Autogenerated Service Key ID
```
alexmensch, line 44: For the `id` value, this is the Terraform ID, correct? So this should be "The ID of the Service Key resource" since it refers to the TF ID.

alexmensch (changes requested):
On the documentation, I may have confused myself, but I'm trying to differentiate between the Twingate ID of an entity, which I think we only ever supply as an input value, versus the Terraform ID for the resource or data source export. So just double check on my comments on which ID is which, and then we'll make the language clear.

docs/data-sources/security_policies.md
```diff
@@ -19,12 +19,15 @@ data "twingate_security_policies" "all" {}
...
 
-- `id` (String) The ID of this resource.
-- `security_policies` (Block List) (see [below for nested schema](#nestedblock--security_policies))
+- `id` (String) Security Policy datasource ID.
 
-<a id="nestedblock--security_policies"></a>
+<a id="nestedatt--security_policies"></a>
 ### Nested Schema for `security_policies`
 
 Read-Only:
```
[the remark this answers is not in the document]
vmanilo: it's output iD, but provider gets it from Twingate backend - it's TwingateID

docs/data-sources/service_accounts.md
```diff
@@ -27,15 +27,15 @@ data "twingate_service_accounts" "foo" {
 
 ### Read-Only
 
-- `id` (String) The ID of this resource.
-- `service_accounts` (Block List) List of Service Accounts (see [below for nested schema](#nestedblock--service_accounts))
+- `id` (String) The ID of the data source result
```
vmanilo, line 30: @alexmensch this one is Terraform ID

golangci.yml
```diff
@@ -65,7 +65,14 @@ linters:
     - structcheck
     - varcheck
     - revive
+    - ireturn
```
ekampf, line 68: Why all these new ignores?

docs/data-sources/resource.md
```diff
@@ -38,16 +38,16 @@ data "twingate_resource" "foo" {
 Read-Only:
 
 - `allow_icmp` (Boolean) Whether to allow ICMP (ping) traffic
-- `tcp` (Block List) (see [below for nested schema](#nestedblock--protocols--tcp))
-- `udp` (Block List) (see [below for nested schema](#nestedblock--protocols--udp))
+- `tcp` (Block, Read-only) (see [below for nested schema](#nestedblock--protocols--tcp))
```
ekampf, line 41: Why is there a change (`Read-only`) here and not in `allow_icmp` ?

docs/data-sources/resources.md
```diff
@@ -30,44 +30,44 @@ data "twingate_resources" "foo" {
 
 ### Read-Only
 
-- `id` (String) The ID of this resource.
-- `resources` (Block List) List of Resources (see [below for nested schema](#nestedblock--resources))
+- `id` (String) The ID of the data source result
```
ekampf, line 33: Why did this change?

docs/data-sources/connectors.md
```diff
@@ -19,17 +19,20 @@ data "twingate_connectors" "all" {}
...
+- `id` (String) The ID of the data source result
 
-<a id="nestedblock--connectors"></a>
+<a id="nestedatt--connectors"></a>
 ### Nested Schema for `connectors`
 
 Read-Only:
 
-- `id` (String) The ID of the Connector.
+- `id` (String) The ID of the Connector resource.
```
ekampf, line 35: why did this change?

docs/data-sources/resource.md
```diff
@@ -29,7 +29,7 @@ data "twingate_resource" "foo" {
 
 - `address` (String) The Resource's address, which may be an IP address, CIDR range, or DNS address
 - `name` (String) The name of the Resource
-- `protocols` (Block List) By default (when this argument is not defined) no restriction is applied, and all protocols and ports are allowed. (see [below for nested schema](#nestedblock--protocols))
+- `protocols` (Block, Read-only) By default (when this argument is not defined) no restriction is applied, and all protocols and ports are allowed. (see [below for nested schema](#nestedblock--protocols))
```
ekampf, line 32: they're all read-only so why change only here?

twingate/internal/provider/datasource/groups.go
```diff
@@ -8,83 +8,102 @@ import (
...
 				Description: "Returns only Groups matching the specified state.",
 			},
-			attr.Type: {
-				Type:         schema.TypeString,
-				Optional:     true,
-				Description:  fmt.Sprintf("Returns only Groups of the specified type (valid: `%s`, `%s`, `%s`).", model.GroupTypeManual, model.GroupTypeSynced, model.GroupTypeSystem),
-				ValidateFunc: validation.StringInSlice([]string{model.GroupTypeManual, model.GroupTypeSynced, model.GroupTypeSystem}, false),
+			attr.Type: schema.StringAttribute{
+				Optional:    true,
+				Description: fmt.Sprintf("Returns only Groups of the specified type (valid options: %s).", utils.DocList(model.GroupTypes)),
```
ekampf, line 78: Why did you make this change?   this doesnt seem related to the move to new sdk

vmanilo: it's just to simplify docs updates, I've notice that we using in multiple places such list of enums
so when we need to add new entry we can only change list at one place

twingate/internal/model/group.go
```diff
@@ -8,6 +8,8 @@ const (
 	GroupTypeSystem = "SYSTEM"
 )
 
+var GroupTypes = []string{GroupTypeManual, GroupTypeSynced, GroupTypeSystem}
```
ekampf, line 11: Seems like an unrelated change

docs/data-sources/connectors.md
vmanilo: @alexmensch asked to change like this https://github.com/Twingate/terraform-provider-twingate/pull/328#discussion_r1235479014

docs/data-sources/resource.md
vmanilo: in the schema we have now separately list of attributes and list of blocks
and for the block we can't set Required/Optional/Computed, this available only for attribute fields

this additional `Read-only` add by doc generator, I'm not setting it 

```
resp.Schema = schema.Schema{
		Description: "Resources in Twingate represent any network destination address that you wish to provide private access to for users authorized via the Twingate Client application. Resources can be defined by either IP or DNS address, and all private DNS addresses will be automatically resolved with no client configuration changes. For more information, see the Twingate [documentation](https://docs.twingate.com/docs/resources-and-access-nodes).",
		Attributes: map[string]schema.Attribute{
			attr.ID: schema.StringAttribute{
				Required:    true,
				Description: "The ID of the Resource. The ID for the Resource can be obtained from the Admin API or the URL string in the Admin Console.",
			},
			// computed
			attr.Name: schema.StringAttribute{
				Computed:    true,
				Description: "The name of the Resource",
			},
			attr.Address: schema.StringAttribute{
				Computed:    true,
				Description: "The Resource's address, which may be an IP address, CIDR range, or DNS address",
			},
			attr.RemoteNetworkID: schema.StringAttribute{
				Computed:    true,
				Description: "The Remote Network ID that the Resource is associated with. Resources may only be associated with a single Remote Network.",
			},
		},
		Blocks: map[string]schema.Block{
			attr.Protocols: schema.SingleNestedBlock{
				Description: "By default (when this argument is not defined) no restriction is applied, and all protocols and ports are allowed.",
				Attributes: map[string]schema.Attribute{
					attr.AllowIcmp: schema.BoolAttribute{
						Computed:    true,
						Description: "Whether to allow ICMP (ping) traffic",
					},
				},
				Blocks: map[string]schema.Block{
					attr.TCP: &protocolSchema,
					attr.UDP: &protocolSchema,
				},
			},
		},
	}
```

docs/data-sources/resource.md
vmanilo: it's the same situation as in previous comment

docs/data-sources/resources.md
vmanilo: @alexmensch proposed to change it like this for all Terraform IDs that we need to set manually for datasources
https://github.com/Twingate/terraform-provider-twingate/pull/328#discussion_r1235470296

vmanilo:
completed in other PRs

Issue #327, opened 2023-05-09
Exception when fetching Group's users


ekampf:
Error:
```
│ Error: failed to read group with id R3JvdXA6NzQ1MjQ=: Message: 400 Bad Request; body: "{\"errors\":[{\"message\":\"Variable '$usersEndCursor' is never used in operation 'readGroup'.\",\"locations\":[{\"line\":1,\"column\":24}],\"path\":null}]}", Locations: [], Extensions: map[code:request_error]
│ 
│   with twingate_group.developers,
│   on twingate_resources_developers.tf line 8, in resource "twingate_group" "developers":
│    8: resource "twingate_group" "developers" {
```

TF Code:
```
locals {
	dev_resources=yamldecode(file("${path.root}/../yaml/dev_resources.yaml"))
	shared_resources=yamldecode(file("${path.root}/../yaml/shared_resources.yaml"))
}

resource "twingate_group" "developers" {
	name ="VPN - Dev Full"
}

resource "twingate_resource" "developers" {
	for_each=merge(local.shared_resources, local.dev_resources)
	name ="Developers - ${each.key}"
	address =each.value.address
	remote_network_id =each.value.environment=="production"?data.twingate_remote_network.production.id : data.twingate_remote_network.test.id
	is_visible =each.value.visible==null?false : true

	protocols {
		allow_icmp =false
		tcp {
			policy ="RESTRICTED"
			ports =each.value.ports
		}
		udp {
			policy ="RESTRICTED"
		}
	}
	access {
		group_ids =[twingate_group.developers.id]
	}
}
```

PR #330, opened 2023-05-11
Fix: added missing cursor

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/327

## Changes
- added missing cursor


ekampf (commented):
Missing unit test?

Issue #338, opened 2023-05-24
Add correlation-id to a TF provider session run


ekampf:
When terraform provider starts (or loads) make a random string (guid?) and send it as header on all outgoing graphql requests.
Header name is `X-Correlation-Id`

PR #341, opened 2023-05-24
Added correlation id

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/338

## Changes
- added header to all requests with correlation id

### DIFF
--- a/twingate/internal/client/client.go
+++ b/twingate/internal/client/client.go
@@ -20,11 +20,13 @@ import (
 )
 
 const (
-	EnvPageLimit = "TWINGATE_PAGE_LIMIT"
-	EnvAPIToken  = "TWINGATE_API_TOKEN" //#nosec
+	EnvPageLimit     = "TWINGATE_PAGE_LIMIT"
+	EnvAPIToken      = "TWINGATE_API_TOKEN" //#nosec
+	EnvCorrelationID = "TWINGATE_CORRELATION_ID"
 
-	headerAPIKey = "X-API-KEY"
-	headerAgent  = "User-Agent"
+	headerAPIKey        = "X-API-KEY"
+	headerAgent         = "User-Agent"
+	headerCorrelationID = "X-Correlation-Id"
 
 	defaultPageLimit = 50
 )
@@ -46,12 +48,14 @@ type Client struct {
 	APIServerURL     string
 	version          string
 	pageLimit        int
+	correlationID    string
 }
 
 type transport struct {
 	underlineRoundTripper http.RoundTripper
 	apiToken              string
 	version               string
+	correlationID         string
 }
 
 func (t *transport) RoundTrip(req *http.Request) (*http.Response, error) {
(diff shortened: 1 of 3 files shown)


twingate/provider.go
```diff
@@ -19,13 +21,19 @@ const (
...
+	EnvNetwork       = "TWINGATE_NETWORK"
+	EnvURL           = "TWINGATE_URL"
+	EnvHTTPTimeout   = "TWINGATE_HTTP_TIMEOUT"
+	EnvHTTPMaxRetry  = "TWINGATE_HTTP_MAX_RETRY"
+	EnvCorrelationID = "TWINGATE_CORRELATION_ID"
 )
 
+func init() { //nolint:gochecknoinits
+	correlationID, _ := uuid.GenerateUUID()
+	_ = os.Setenv(EnvCorrelationID, correlationID)
```
ekampf, line 34: Why do we have different parts of our code communicating a value via OS env var?

This is not a good way and allows interfering with it from the outside

vmanilo:
Initially I tried to init this value directly in http client constructor, but when I run some acc test, I’ve noticed that this id was changing from request to request.I can test some other approach as well to not use env var, but even if someone defines the same env var we will overwrite it.25 мая 2023 г., в 5:40 AM, Eran Kampf ***@***.***> написал(а):﻿
@ekampf commented on this pull request.



In twingate/provider.go:
>  )
 
+func init() { //nolint:gochecknoinits
+	correlationID, _ := uuid.GenerateUUID()
+	_ = os.Setenv(EnvCorrelationID, correlationID)

Why do we have different parts of our code communicating a value via OS env var?
This is not a good way and allows interfering with it from the outside

twingate/provider.go
vmanilo: Initially I tried to init this value directly in http client constructor, but when I run some acc test, I’ve noticed that this id was changing from request to request.I can test some other approach as well to not use env var, but even if someone defines the same env var we will overwrite it.

vmanilo: I will be back online in a few hours, and can try another approach

vmanilo: ok, changed to global var

twingate/internal/client/client.go
```diff
@@ -16,15 +16,19 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-retryablehttp"
+	"github.com/hashicorp/go-uuid"
 	"github.com/hasura/go-graphql-client"
 )
 
+var correlationID, _ = uuid.GenerateUUID() //nolint:gochecknoglobals
```
ekampf, line 23: Why do we need a global var (and a new lint ignore) when we can just do this in `NewClient` ?
When you create a new client just send it `correlationID: uuid.GenerateUUID()`

vmanilo: fixed

ekampf (approved):

PR #349, opened 2023-06-20
Dont Smoke-Test terraform versions no longer supported

ekampf:
## Changes
- `latest` moved to be 1.5 but we still want to tst on 1.4. So this explicitly lists all versions we currently support

Screenshot 2023-06-20 at 1 06 57 PM


romankor (approved):
Gtm

PR #354, opened 2023-07-02
Upgrade TF provider - part 1

vmanilo:
Resolves upgrading provider to new TF plugin framework

Test passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/5731610432/job/15532879663

## Changes
- setup provider in mix mode


twingate/provider.go
```diff
@@ -63,56 +65,57 @@ func Provider(version string) *schema.Provider {
 func providerOptions() map[string]*schema.Schema {
 	return map[string]*schema.Schema{
 		attr.APIToken: {
-			Type:        schema.TypeString,
-			Optional:    true,
-			Sensitive:   true,
-			DefaultFunc: schema.EnvDefaultFunc(EnvAPIToken, nil),
```
ekampf, line 69: Note to self - we remove default from schema because new plugins sdk doesnt support it.
So these default values moved to `configure` func

PR #387, opened 2023-08-18
Upgrade Resource datasource - part 11

vmanilo:
Resolves upgrade TF provider

Test passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/5906306390?pr=26

## Changes
- upgrade datasource Resource
- remove provider mix mode


docs/data-sources/resource.md
```diff
@@ -38,16 +38,16 @@ data "twingate_resource" "foo" {
 Read-Only:
 
 - `allow_icmp` (Boolean) Whether to allow ICMP (ping) traffic
```
ekampf, line 40: How come no readonly here?

vmanilo: in datasource mostly all fields are computed, I think that's why it listed in ReadOnly section

vmanilo: oh did you mean, there is no readonly written on this field?

we don't control this, thats doc generation library adds this extra info, maybe for simple fields it's omitted 
as I see the same for: name, address, policy

Issue #407, opened 2023-10-04
Fix dependency between service accounts and service keys when applying changes


alexmensch:
Below are details on where I ran into this problem in a Terraform configuration. Let me know if you need more information.

`terraform plan` wanted to leave a service key in place, and then change the parent service account, but that’s actually not allowed by Twingate.

I got this error on run: `Error: failed to delete service account with id abcde: Github CI/CD contains 1 active keys that must be revoked first.`

And this is what plan said:

```Terraform will perform the following actions:

  # module.cluster.twingate_resource.cluster will be updated in-place
  ~ resource "twingate_resource" "cluster" {
        id                          = "UmVzb3VyY2U6MjMzNzUzOA=="
        name                        = "GKE Cluster"
        # (6 unchanged attributes hidden)

      ~ access {
          ~ service_account_ids = [] -> (known after apply)
            # (1 unchanged attribute hidden)
        }

        # (1 unchanged block hidden)
    }

  # module.cluster.twingate_service_account.github_actions will be created
  + resource "twingate_service_account" "github_actions" {
      + id   = (known after apply)
      + name = "Github CI/CD"
    }

  # module.gitops.twingate_service_account.github_actions will be destroyed
  # (because twingate_service_account.github_actions is not in configuration)
  - resource "twingate_service_account" "github_actions" {
      - id   = "U2VydmljZUFjY291bnQ6MTJkZmFhMjQtNTQyYi00YzZjLTg5YWUtZWEyMmIzMmMzYzJj" -> null
      - name = "Github CI/CD" -> null
    }

  # module.gitops.twingate_service_account_key.github_actions_key will be updated in-place
  ~ resource "twingate_service_account_key" "github_actions_key" {
        id                 = "U2VydmljZUFjY291bnRLZXk6c0Y2T0VRdFphRE83TkF6MWtOdVo1c2ktR3YzR25TZklDUHNlTGRZVU9Baw=="
        name               = "rustling-scorpion"
      ~ service_account_id = "U2VydmljZUFjY291bnQ6MTJkZmFhMjQtNTQyYi00YzZjLTg5YWUtZWEyMmIzMmMzYzJj" -> (known after apply)
        # (1 unchanged attribute hidden)
    }

Plan: 1 to add, 2 to change, 1 to destroy.
```

So what we should be doing here is destroying the key, too, and then recreating the key and service account per the changes in the configuration.

To fix this, I have to manually delete the service account key.

[...]

PR #415, opened 2023-10-19
Fix: service account and service key

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/407

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/6569760290/job/17846156883

## Changes
- fix by revoking service_key before deleting service_account

### DIFF
--- a/twingate/internal/client/service-account.go
+++ b/twingate/internal/client/service-account.go
@@ -72,19 +72,6 @@ func (client *Client) UpdateServiceAccount(ctx context.Context, serviceAccount *
 }
 
 func (client *Client) DeleteServiceAccount(ctx context.Context, serviceAccountID string) error {
-	serviceAccount, err := client.ReadServiceAccount(ctx, serviceAccountID)
-	if err != nil && !errors.Is(err, ErrGraphqlResultIsEmpty) {
-		return err
-	}
-
-	if serviceAccount != nil {
-		for _, key := range serviceAccount.Keys {
-			if err := client.RevokeServiceKey(ctx, key); err != nil {
-				return err
-			}
-		}
-	}
-
 	opr := resourceServiceAccount.delete()
 
 	if serviceAccountID == "" {
--- a/twingate/internal/client/service-account.go
+++ b/twingate/internal/client/service-account.go
@@ -72,6 +72,19 @@ func (client *Client) UpdateServiceAccount(ctx context.Context, serviceAccount *
 }
 
 func (client *Client) DeleteServiceAccount(ctx context.Context, serviceAccountID string) error {
+	serviceAccount, err := client.ReadServiceAccount(ctx, serviceAccountID)
+	if err != nil && !errors.Is(err, ErrGraphqlResultIsEmpty) {
+		return err
+	}
+
+	if serviceAccount != nil {
+		for _, key := range serviceAccount.Keys {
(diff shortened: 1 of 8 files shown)


twingate/internal/client/service-account.go
```diff
@@ -72,6 +72,19 @@ func (client *Client) UpdateServiceAccount(ctx context.Context, serviceAccount *
...
 
 func (client *Client) DeleteServiceAccount(ctx context.Context, serviceAccountID string) error {
+	serviceAccount, err := client.ReadServiceAccount(ctx, serviceAccountID)
+	if err != nil && !errors.Is(err, ErrGraphqlResultIsEmpty) {
+		return err
+	}
+
+	if serviceAccount != nil {
+		for _, key := range serviceAccount.Keys {
+			if err := client.RevokeServiceKey(ctx, key); err != nil {
```
ekampf, line 82: Not sure I understand the process here.
The Service Account __owns__ its keys so if someone deleted it - the keys are gone too. 
There's no need to make API calls to revoke them...

ekampf: Also mutation would error in this case...

twingate/internal/client/service-account.go
```diff
@@ -72,6 +72,19 @@ func (client *Client) UpdateServiceAccount(ctx context.Context, serviceAccount *
...
+				return err
+			}
+		}
+	}
+
 	opr := resourceServiceAccount.delete()
```
ekampf, line 88: So before this PR, we ignores the SA's keys and 
- Immediately deleted it from TF state
- Called `DeleteServiceAccount` mutation - which would have returned an error, because you cannot delete an SA with valid keys.

So it seems to me the problem here that TF state gets updated regardless to the result of the delete mutation?

ekampf: Seems like an issue with all the delete methods?  (`DeleteGroup`, `DeleteResource` etc seem to behave the same)

twingate/internal/client/service-account.go
```diff
@@ -72,6 +72,19 @@ func (client *Client) UpdateServiceAccount(ctx context.Context, serviceAccount *
 }
 
 func (client *Client) DeleteServiceAccount(ctx context.Context, serviceAccountID string) error {
+	serviceAccount, err := client.ReadServiceAccount(ctx, serviceAccountID)
+	if err != nil && !errors.Is(err, ErrGraphqlResultIsEmpty) {
+		return err
+	}
+
+	if serviceAccount != nil {
```
ekampf, line 80: I dont understand this fix and the relation to #407.
This code is correct - if we want TF to delete an SA, it has to revoke all its keys first.
But what happened in the issue is something else - the SA was deleted outside of TF:
- In this case `serviceAccount` is `nil` and this block doesnt even run
- `DeleteServiceAccount` mutation will succeed -> we'll delete the SA from TF state **but not** its keys

ekampf: We need a reproduction unit tests that:
1. Creates an SA + key using TF
2. Makes "manual" API calls to revoke the key and delete the SA
3. Checked TF re-creates both SA + key

vmanilo: updated PR

ekampf (approved):

Issue #430, opened 2023-12-08
Add support for disabling resources in twingate_resource


twingate-blee:
### New functionality

- (Optional) `is_active` of type `Boolean`
  - Set the resource as active or inactive. Default is `true`

Behavior

- Setting `is_active` to false for a Resource will set the resource as disabled in the console.
![Screenshot 2023-12-08 at 1 47 56 PM](https://github.com/Twingate/terraform-provider-twingate/assets/56551241/eedb7be7-ab48-4652-82e8-b86062ae468c)

Example usage

```terraform
resource "twingate_resource" "internal_resources" {
  name = "network"
  address = "internal.int"
  remote_network_id = twingate_remote_network.aws_network.id
  is_active = false
}
```

### Bug
When a resource is disabled from the console when you run Terraform `plan` it will enable the resource.

This seems to run on a plan as well as an apply https://github.com/Twingate/terraform-provider-twingate/blob/6c2e491543acbd9119f9cf3b52df07f4a2b4e81b/twingate/internal/provider/resource/resource.go#L329

PR #438, opened 2023-12-11
Added optional attribute is_active to twingate_resource

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/430

Test passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/7172949273

## Changes
- added optional attribute `is_active` to twingate_resource


twingate-blee (approved):
lgtm

twingate/internal/provider/resource/resource.go
```diff
@@ -185,6 +192,10 @@ func resourceCreate(ctx context.Context, resourceData *schema.ResourceData, meta
 		return diag.FromErr(err)
 	}
 
+	if !resource.IsActive {
+		return diag.FromErr(ErrResourceInactiveOnCreation)
```
tjstansell, line 196: Just curious, why can't the provider handle this for us and just create the resource and then set it to inactive immediately to reduce friction for the end user? Seems odd to me that the API restricts this, TBH.

vmanilo: thank you for your comment, it's fixed now

so now user can set inactive flag from the beginning

Issue #436, opened 2023-12-08
Add support for filtering for twingate_service_accounts data source


twingate-blee:
**Modified functionality**

Only one of name, name_regex, name_contains, name_exclude, name_prefix or name_suffix must be set.

- (Optional) `name` of type `string`
  - Returns only service accounts that exactly match this name.
 
- (Optional) `name_regex` of type `string`
  - The regular expression match of the name of the service account.

- (Optional) `name_contains` of type `string`
  - Match when the value exist in the name of the service account.

- (Optional) `name_exclude` of type `string`
  - Match when the value does not exist in the name of the service account.
 
- (Optional) `name_prefix` of type `string`
  - The name of the service account must start with the value.

- (Optional) `name_suffix` of type `string`
  - The name of the service account must end with the value.

**Behavior**

- Current behavior returns a service account that's an exact match by name. This will allow filtering based on the criteria above

**Example usage**

```terraform
data "twingate_service_accounts" "filter" 
{
  name_contains = "test"
}
```

PR #443, opened 2023-12-18
Feature: add filtering support for twingate resources datasource

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/434 https://github.com/Twingate/terraform-provider-twingate/issues/356

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/7306475547

## Changes
- added optional attributes to resources datasource

PR #445, opened 2023-12-24
Feature: add filtering for service accounts datasource

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/436

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/7312834045

## Changes
- added optional name filters


docs/data-sources/resources.md
```diff
@@ -15,6 +15,11 @@ Resources in Twingate represent servers on the private network that clients can
 ```terraform
 data "twingate_resources" "foo" {
   name = "<your resource's name>"
+#  name_regexp = "<your resource's name>"
```
twingate-blee, line 18: This file should probably be removed and the one from https://github.com/Twingate/terraform-provider-twingate/pull/443/files#diff-4cf56f1941dcb68e635ce12e6ddc09fe0ca9d321a15dec471543f27ed15f4e5b should be used instead

docs/data-sources/service_accounts.md
```diff
@@ -23,7 +23,12 @@ data "twingate_service_accounts" "foo" {
 
 ### Optional
 
-- `name` (String) Filter results by the name of the Service Account.
```
twingate-blee, line 26: We should probably put some wording in here that if no options are passed it will return all service accounts. Also only one option can be used at a time.

examples/data-sources/twingate_resources/data-source.tf
```diff
@@ -1,5 +1,10 @@
 data "twingate_resources" "foo" {
   name = "<your resource's name>"
+#  name_regexp = "<your resource's name>"
```
twingate-blee, line 3: Update with
```
#  name_regexp = "<regular expression of resource name>"
#  name_contains = "<a string in the resource name>"
#  name_exclude = "<your resource's name to exclude>"
#  name_prefix = "<prefix of resource name>"
#  name_suffix = "<suffix of resource name>"
```

twingate-blee:
Functionality works great. Just a few requested changes for the docs.

docs/data-sources/resources.md
vmanilo: fixed

docs/data-sources/service_accounts.md
vmanilo: sure, added

examples/data-sources/twingate_resources/data-source.tf
vmanilo: updated

twingate-blee (approved):
lgtm

PR #447, opened 2024-01-07
Fix: datasource twingate groups with empty filters

vmanilo:
Resolves failing acc tests

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/7440531634/job/20241637660

## Changes
- fix acc test datasource twingate groups with empty filters


twingate/internal/test/acctests/datasource/groups_test.go
```diff
@@ -176,22 +176,31 @@ func TestAccDatasourceTwingateGroups_WithEmptyFilters(t *testing.T) {
 			},
 			Steps: []resource.TestStep{
 				{
-					Config:             testTwingateGroupsWithEmptyFilter(),
+					Config:             testTwingateGroupsWithEmptyFilter(test.RandomGroupName()),
 					ExpectNonEmptyPlan: true,
+					Check: acctests.ComposeTestCheckFunc(
+						testCheckResourceAttrNotEqual("data.twingate_groups.all", groupsLen, "0"),
```
ekampf, line 182: Why would it be empty with no filters?  It should return all...

vmanilo: I'm expecting at least one group, but we might have more on the test tenant

ekampf (approved):

PR #453, opened 2024-01-16
Fix: create twingate-connector with disabled status

vmanilo:
Resolves issue with creating new connector with disabled status

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/7547395908

## Changes
- passing optional parameter `status_updates_enabled` on creation
- added acc test


romankor (approved):
Gtm

docs/resources/connector.md
```diff
@@ -24,6 +24,7 @@ resource "twingate_remote_network" "aws_network" {
```
twingate-blee, line 24: Can you update the doc to show what the default value is for `status_updates_enabled`

twingate-blee: I made the change

Issue #437, opened 2023-12-08
Add support for filtering for twingate_users data source


twingate-blee:
**Modified functionality**

Only one of email, email_regex, email_contains, email_exclude, email_prefix or email_suffix must be set.

- (Optional) `email` of type `string`
  - Returns only users that exactly match this email.
 
- (Optional) `email_regex` of type `string`
  - The regular expression match of the email of the user.

- (Optional) `email_contains` of type `string`
  - Match when the value exist in the email of the user.

- (Optional) `email_exclude` of type `string`
  - Match when the value does not exist in the email of the user.
 
- (Optional) `email_prefix` of type `string`
  - The email of the user must start with the value.

- (Optional) `email_suffix` of type `string`
  - The email of the user must end with the value.

Only one of first_name, first_name_regex, first_name_contains, first_name_exclude, first_name_prefix or first_name_suffix must be set.

- (Optional) `first_name` of type `string`
  - Returns only users that exactly match the first name.
 
- (Optional) `first_name_regex` of type `string`
  - The regular expression match of the first name of the user.

- (Optional) `first_name_contains` of type `string`
  - Match when the value exist in the first name of the user.

- (Optional) `first_name_exclude` of type `string`
  - Match when the value does not exist in the first name of the user.
 
- (Optional) `first_name_prefix` of type `string`
  - The first name of the user must start with the value.

- (Optional) `first_name_suffix` of type `string`
  - The first name of the user must end with the value.

Only one of last_name, last_name_regex, last_name_contains, last_name_exclude, last_name_prefix or last_name_suffix must be set.

- (Optional) `last_name` of type `string`
  - Returns only users that exactly match the last name.
 
- (Optional) `last_name_regex` of type `string`
  - The regular expression match of the last name of the user.

- (Optional) `last_name_contains` of type `string`
  - Match when the value exist in the last name of the user.

- (Optional) `last_name_exclude` of type `string`
  - Match when the value does not exist in the last name of the user.
 
- (Optional) `last_name_prefix` of type `string`
  - The last name of the user must start with the value.

- (Optional) `last_name_suffix` of type `string`
  - The last name of the user must end with the value.

- (Optional) `roles` of type `[string]`
  - Returns users that match a list of roles. valid roles : `ADMIN `, `DEVOPS `, `SUPPORT `, `MEMBER`.
      
**Behavior**

- Current behavior returns all users. This will allow filtering based on the criteria above

**Example usage**

```terraform
[...]

PR #456, opened 2024-01-18
[After #443] Feature: add filtering support for twingate_users datasource

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/437

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/7574865850

## Changes
- added optional filters


twingate-blee (approved):
lgtm

Issue #433, opened 2023-12-08
Add support for filtering for twingate_remote_networks data source


twingate-blee:
**Modified functionality**

Only one of name, name_regex, name_contains, name_exclude, name_prefix or name_suffix must be set.

- (Optional) `name` of type `string`
  - Returns only remote networks that exactly match this name.
 
- (Optional) `name_regex` of type `string`
  - The regular expression match of the name of the remote network.

- (Optional) `name_contains` of type `string`
  - Match when the value exist in the name of the remote network.

- (Optional) `name_exclude` of type `string`
  - Match when the value does not exist in the name of the remote network.
 
- (Optional) `name_prefix` of type `string`
  - The name of the remote network must start with the value.

- (Optional) `name_suffix` of type `string`
  - The name of the remote network must end with the value.

**Behavior**

- Current behavior returns all remote networks. This will allow filtering based on the criteria above

**Example usage**

```terraform
data "twingate_remote_networks" "filter" 
{
  name_regex = "test.*"
}
```

PR #458, opened 2024-01-22
[After #443] Feature: add filtering for twingate remote networks datasource

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/433

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/7607118051/job/20713942883

## Changes
- added filter options for remote-networks datasource


docs/data-sources/remote_networks.md
```diff
@@ -19,6 +19,15 @@ data "twingate_remote_networks" "all" {}
 <!-- schema generated by tfplugindocs -->
 ## Schema
 
+### Optional
```
twingate-blee, line 22: Should we add commented examples to "Example Usage" similar to the other resources?

vmanilo: yes, added example

Issue #459, opened 2024-01-24
twingate_resource update failing to remove service_account


kaarla:
I'm updating a twingate_resource, removing a service account access I don't need anymore, this is my code:

```
resource "twingate_resource" "snowflake" {
  name                        = "Snowflake"
  address                     = "[redacted]"
  remote_network_id           = data.twingate_remote_network.[redacted].id
  is_browser_shortcut_enabled = true

  protocols = {
    allow_icmp = true
    tcp = {
      policy = "ALLOW_ALL"
    }
    udp = {
      policy = "ALLOW_ALL"
    }
  }

  access {
    group_ids = [twingate_group.snowflake.id]
  }
}
``` 

The first time, the plan it showed correctly that the access would be removed:
```
update
access {
delete
service_account_ids :
[
delete
"[redacted]"
]
null
}
```

It successfully removed the access, but retrieves this error after applying:

```
Error: Provider produced inconsistent result after apply
When applying changes to twingate_resource.snowflake, provider "provider[\"registry.terraform.io/twingate/twingate\"]" produced an unexpected new value: .access[0].service_account_ids: was null, but now cty.SetVal([]cty.Value{cty.StringVal("[redacted]")}).

This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

I tried running a new plan, but even now that the account access is not present anymore, plans to remove it again instead having an empty plan. I tried assigning back the service account manually and triggered a new plan but the result is the same.

twingate-blee:
@kaarla Which version of the Twingate provider are you using? I think this is now fixed in v2.0.0 that was released last week

kaarla:
@twingate-blee I was using `v2.0.0` implicitly, now I tried using it explicitly just in case, but the result is the same, when it plans still finds the service account that is not assigned anymore.

Code:
```
terraform {
  required_providers {
    aws = {
      source  = "hashicorp/aws"
      version = "~> 4.0"
    }
    twingate = {
      source = "twingate/twingate"
      version = "~> 2.0.0"
    }
  }
}
```

Plan:

![Screenshot 2024-01-24 at 11 56 03 a m](https://github.com/Twingate/terraform-provider-twingate/assets/29378995/4c3d5dd1-3eb8-4f93-adcb-21e591877661)

Infrastructure (Snowflake Access is a group, not a service account):

[...]

PR #461, opened 2024-01-26
Fix: resource access

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/459

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/7662687464/job/20884332505

## Changes
- fixed updating access block when completely removing groups or service account ids


twingate-blee (approved):
I was able to duplicate the issue at https://github.com/Twingate/terraform-provider-twingate/issues/459

I then updated to the fixed version of the provider and no longer go the error or was able to reproduce it again.

Issue #432, opened 2023-12-08
Add support for filtering for twingate_groups data source


twingate-blee:
**Modified functionality**

Only one of name, name_regex, name_contains, name_exclude, name_prefix, name_suffix, or types must be set.

- (Optional) `name` of type `string`
  - Returns only groups that exactly match this name.
 
- (Optional) `name_regex` of type `string`
  - The regular expression match of the name of the group.

- (Optional) `name_contains` of type `string`
  - Match when the value exist in the name of the group.

- (Optional) `name_exclude` of type `string`
  - Match when the value does not exist in the name of the group.
 
- (Optional) `name_prefix` of type `string`
  - The name of the group must start with the value.

- (Optional) `name_suffix` of type `string`
  - The name of the group must end with the value.

Update the existing type with types
- (Optional) `types` of type `[string]`
  - Returns groups that match a list of types. valid types : `MANUAL`, `SYNCED`, `SYSTEM`.

**Behavior**

- Current behavior returns a group that's an exact match by name. This will allow filtering based on the criteria above

**Example usage**

```terraform
data "twingate_groups" "filter" 
{
  types = ["SYNCED","SYSTEM"]
}
```

PR #462, opened 2024-01-27
[After #443] Feature: add support for filtering twingate groups datasource

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/432

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/7676594341

## Changes
- added optional filters for groups datasource


docs/data-sources/groups.md
```diff
@@ -28,8 +33,13 @@ data "twingate_groups" "foo" {
 
 - `groups` (Attributes List) List of Groups (see [below for nested schema](#nestedatt--groups))
```
twingate-blee, line 34: Is there a difference between

```
data "twingate_groups" "foo" {
  groups = {
    name = "test"
    is_active = true
  }
}  
``` 

and

```
data "twingate_groups" "foo" {
  name = "test"
  is_active = true
}  
```

Do we still need the `groups` attribute? If so should we just add the filters to it?

ekampf: We dont have this extra hierarchy in other datasources?
Personally, I would rather have a breaking change and get rid of it  than have this weird duplication of functionality

ekampf: I see we have this pattern in all our datasources...

vmanilo: fixed

twingate-blee (approved):
lgtm

PR #463, opened 2024-01-27
[After #443] Feature: add support filtering twingate_connectors datasource

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/431

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/7677158805/job/20925569374

## Changes
- added optional filters to twingate_connectors datasource


docs/data-sources/connectors.md
```diff
@@ -22,6 +29,12 @@ data "twingate_connectors" "all" {}
 ### Optional
 
 - `connectors` (Attributes List) List of Connectors (see [below for nested schema](#nestedatt--connectors))
```
twingate-blee, line 31: Is there a difference between

```
data "twingate_connectors" "all" {
  connectors = {
    name = "test"
  }
}  
``` 

and

```
data "twingate_connectors" "all" {
  name = "test"
}  
```

Do we still need the `connectors` attribute? If so should we just add the filters to it?

twingate-blee, line 31: The `connectors` attribute is not an input. This type needs to be updated so it will be moved to the "Read-Only" section in the docs

vmanilo: fixed

Issue #422, opened 2023-11-07
Add support for policies in Resource access blocks


alexb-twingate:
Modified functionality

- (Optional) `access_group` block
  - `group_id` - a single group ID can be passed to each access block
  - (Optional) `security_policy_id` of type `string`
    - The ID of a `twingate_security_policy` to use as the access policy for the group IDs in the access block
- (Optional) `access_service` block
  - `service_account_id` - the ID of the service account that should have access to this Resource.

Behavior

- Setting `security_policy_id` Inside of an `access_group` block will set the Policy that's used to access the Resource by the `group_ids` inside of the `access` block.
  - If `security_policy_id` isn't specified of null, then the access will use the Resource's Policy.
- `access_service` is now used to assign a service account access to a Resource.
  -  `security_policy_id` field cannot be used in `access_service` blocks. Service accounts do not use security policies and trying to set a policy on an access between a Resource and a service account will throw an error.
- As with #245, when `is_authoritative` is true, all `access` blocks will override any existing assignments. If false, assignments made outside of Terraform will be ignored.

Example usage

```terraform
data "twingate_groups" "devops" {
  name = "DevOps"
}

data "twingate_groups" "security" {
  name = "Security"
}

data "twingate_security_policy" "mfa" {
  name = "MFA"
}

data "twingate_security_policy" "no_mfa" {
  name = "No MFA"
}

resource "twingate_service_account" "github_actions_prod" {
  name = "Github Actions PROD"
}

resource "twingate_remote_network" "aws" {
  name = "AWS"
}

resource "twingate_resource" "internal_resources" {
  name = "All internal resources"
  address = "*.internal.int"
  remote_network_id = twingate_remote_network.aws.id

  security_policy_id = twingate_security_policy.mfa.id

  protocols = {
    allow_icmp = true
    tcp  {
      policy = "RESTRICTED"
      ports = ["80", "82-83"]
    }
    udp {
      policy = "ALLOW_ALL"
    }
  }

  access_group {
    security_policy_id = twingate_security_policy.no_mfa.id
    group_id = data.twingate_groups.devops.id
  }
 
  // Alternatively to specify multiple groups:
  dynamic access_group {
    for_each = toset([twingate_groups.devops.id, twingate_groups.security.id])
    content {
      security_policy_id = twingate_security_policy.no_mfa.id
      group_id = access.value.key
    }
  }
  
  // Service accounts are now assigned via the `service_access` block
  access_service {
[...]

PR #500, opened 2024-04-01
Feature: add support for policies in resource access blocks

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/422

Acc tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/8513637200
unit test had only issue with linter, which fixed in last commit

## Changes
- breaking change for Resource access block - divided onto two blocks access_group and access_service


twingate-blee (approved):
I converted my project from using access to access_group. I also tested access_service and added and removed security_policy_id

docs/resources/resource.md
```diff
@@ -52,9 +52,19 @@ resource "twingate_resource" "resource" {
     }
   }
 
-  access {
-    group_ids = [twingate_group.aws.id]
-    service_account_ids = [twingate_service_account.github_actions_prod.id]
+  dynamic "access_group" {
```
alexb-twingate, line 55: Can we add an example of a single `access_group` block as well? While it might not be needed, it would be nice to have an explicit example of a single `access_group` block being used.

Issue #499, opened 2024-03-26
Reduce requests for readResource graphql op


bertekintw:
Today TF needs to do drift check so for every TF resource it makes an API call (readResource) and compares server state to its state. For a tenant with 1,000 resources, it means 1000 readResource requests.

We should optimize this:
TF fetches ALL resources in batch (paginate with 100 page size) at start and store in mem
drift checks would run against memory

So for the tenant with 1,000 resources, the number of API requests will be reduced to 10.

PR #516, opened 2024-04-24
Feature: reduce requests for resource read operation

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/499

Tests passes:  https://github.com/vmanilo/terraform-provider-twingate/actions/runs/8865221925

## Changes
- added reading cache


twingate-blee:
Tested this by creating 100 resources and then destroying them.

Blue is the number of calls using caching. Green is without
Screenshot 2024-04-26 at 2 19 19 PM

twingate-blee (approved):
lgtm

PR #517, opened 2024-04-25
Fix schema upgrade

vmanilo:
Resolves issue with upgrading from previous provider version

## Changes
- fix object atribute types


twingate-blee (approved):
This fixed the issue we had on the test tenant

Issue #552, opened 2024-06-25
crash during twingate_resource update for bad alias value


builtroller:
Terraform Version: 1.8.5
twingate provider version: 3.0.5

I was attempting to make a simple change to an existing `twingate_resource`. 
- update the `name`
- add an `alias`

Here is the relevant portion of the plan:
```tf
  # twingate_resource.core80 will be updated in-place
  ~ resource "twingate_resource" "core80" {
      + alias                       = "dev-platform-core-us-east-1"
        id                          = "UmVzb3VyY2U6MjUxMzIxOQ=="
      ~ name                        = "dev-platform-core-us-east-1" -> "db-dev-platform-core-us-east-1"
        # (8 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }
```

And here is the error:
```tf
╷
│ Error: Plugin did not respond
│ 
│ The plugin encountered an error, and failed to respond to the
│ plugin6.(*GRPCProvider).ApplyResourceChange call. The plugin logs may
│ contain more details.
╵

Stack trace from the terraform-provider-twingate_v3.0.5 plugin:

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x86539a]

goroutine 16 [running]:
github.com/Twingate/terraform-provider-twingate/v3/twingate/internal/provider/resource.(*twingateResource).Update(0xc00006a548, {0xdf6fc8, 0xc00034e990}, {{{{0xdfbfd0, 0xc0003bbdd0}, {0xbb7e80, 0xc0003a7470}}, {0xdfe010, 0xc00053a870}}, {{{0xdfbfd0, ...}, ...}, ...}, ...}, ...)
	github.com/Twingate/terraform-provider-twingate/v3/twingate/internal/provider/resource/resource.go:833 +0x63a
github.com/hashicorp/terraform-plugin-framework/internal/fwserver.(*Server).UpdateResource(0xc00032a000, {0xdf6fc8, 0xc00034e990}, 0xc0005ab510, 0xc0005ab4e8)
	github.com/hashicorp/terraform-plugin-framework@v1.9.0/internal/fwserver/server_updateresource.go:122 +0x6ee
github.com/hashicorp/terraform-plugin-framework/internal/fwserver.(*Server).ApplyResourceChange(0xc00032a000, {0xdf6fc8, 0xc00034e990}, 0xc00053a500, 0xc0005ab6d0)
[...]

PR #553, opened 2024-06-26
Fix: crash on invalid alias value

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/552

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/9681717928/job/26712959782

## Changes
- fix crash on nil resource


twingate-blee (approved):
lgtm

PR #566, opened 2024-08-05
Updated group query names

vmanilo:
## Changes
- update group query names

use cases:

1. Group resource
  1. Init cache at beginning of provider: op will be `cache_readGroups` for first call and then `cache_readGroups_readGroupsAfter` for all pages
  2. Create a group and read it after - op name is just `readGroup`
2. Group datasource (singular) - op name is `datasource_readGroup`
3. Group datasource (plural) - op name is `datasource_readGroups` and for pages `datasource_readGroups_readGroupsAfter`

### DIFF
--- a/twingate/internal/client/group.go
+++ b/twingate/internal/client/group.go
@@ -57,7 +57,9 @@ func (client *Client) ReadGroup(ctx context.Context, groupID string) (*model.Gro
 		return nil, err
 	}
 
-	if err := response.Group.Users.FetchPages(ctx, client.readGroupUsersAfter, variables); err != nil {
+	oprCtx := withOperationCtx(ctx, opr)
+
+	if err := response.Group.Users.FetchPages(oprCtx, client.readGroupUsersAfter, variables); err != nil {
 		return nil, err //nolint
 	}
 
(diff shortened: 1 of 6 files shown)


twingate/internal/client/group.go
```diff
@@ -79,33 +81,35 @@ func (client *Client) ReadGroups(ctx context.Context, filter *model.GroupsFilter
...
+	if err := client.query(ctx, &response, variables, opr, attr{id: "All"}); err != nil {
 		return nil, err
 	}
 
 	return &response.PaginatedResource, nil
 }
 
 func (client *Client) ReadFullGroups(ctx context.Context) ([]*model.Group, error) {
-	opr := resourceGroup.read()
+	opr := resourceGroup.read().withCustomName("cache_readGroups")
```
ekampf, line 112: why cache_?

vmanilo: we use this call from client cache

vmanilo: do you prefer `resource_`?

vmanilo: or simply `readFullGroups` ?

ekampf (approved):
Fix usecase 4 and merge

PR #575, opened 2024-08-27
Fix: dns filtering profile tests

vmanilo:
Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/10572136583

## Changes
- fixed DNS filtering resource - read operation


twingate-blee (approved):
Lgtm

PR #580, opened 2024-09-15
Fix/read dns profiles

vmanilo:
Test passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/10869162404

## Changes
- added custom timeout and retry for DNS profiles query

### DIFF
--- a/twingate/internal/client/client.go
+++ b/twingate/internal/client/client.go
@@ -36,6 +36,9 @@ const (
 	extendedPageLimit = 100
 
 	defaultRateLimit = 3
+
+	defaultQueryTimeout = time.Second * 10
+	defaultQueryRetries = 3
 )
 
 var (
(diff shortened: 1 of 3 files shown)


twingate/internal/client/client.go
```diff
@@ -340,6 +343,19 @@ type ResponseWithPayload interface {
 	IsEmpty() bool
 }
 
+func (client *Client) queryWithTimeout(ctx context.Context, resp ResponseWithPayload, variables map[string]any, opr operation, attrs ...attr) error {
+	timeoutCtx, cancel := context.WithTimeout(ctx, defaultQueryTimeout)
```
twingate-blee, line 347: When context.WithTimeout is set outside the retry loop does that mean it has 10 seconds to complete including all retries?

If we set this inside the loop will each retry have 10 seconds to complete?

vmanilo: thank you, fixed it

twingate-blee (approved):
lgtm

PR #594, opened 2024-11-01
Fix: dns filtering profile mutation

vmanilo:
Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/11634455249

## Changes
- added extra retry for DNS filtering profile mutation


twingate-blee (approved):
tested. lgtm

PR #596, opened 2024-11-04
Fix: improve acc tests

vmanilo:
## Changes
- parallelize acc tests


.github/workflows/ci.yml
```diff
@@ -120,14 +120,13 @@ jobs:
...
-    timeout-minutes: 15
     strategy:
-      max-parallel: 2
+      max-parallel: 3
       fail-fast: false
       matrix:
         terraform:
           - '1.8.*'
-          - '1.9.*'
+#          - '1.9.*'
```
twingate-blee, line 129: Will we add this back when Terraform 1.10 comes out?

vmanilo: yes

PR #597, opened 2024-11-04
Fix: added linter versioning

vmanilo:
Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/11673411857

## Changes
- added `golangci-lint` to tools.go and dependencies list


vmanilo:
We decided to go with dockerised version https://github.com/Twingate/terraform-provider-twingate/pull/601

PR #618, opened 2024-12-09
Enable Terraform 1.9.* for CI tests

vmanilo:
Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/12243391198/job/34152892466

## Changes
- added terraform 1.9.*


.github/workflows/ci.yml
```diff
@@ -126,7 +126,7 @@ jobs:
       matrix:
         terraform:
           - '1.8.*'
```
ekampf, line 128: We can remove 1.8 as its EOL

Screenshot 2024-12-10 at 8 55 11 AM

.github/workflows/smoketests.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       matrix:
         terraform:
           - '1.8.*'
```
ekampf, line 48: same

vmanilo: sure, removed it

.github/workflows/ci.yml
vmanilo: good point

twingate-blee (approved):
lgtm

PR #623, opened 2025-01-03
Fix: acoount typo

pcarney8:
## Changes
- there was a typo. it says `acoount` and means `account`


alexb-twingate (approved):
That's silly of us! Thanks for catching this.

PR #637, opened 2025-02-10
Fix connector name

vmanilo:
Resolves issue with empty connector name on pulumi provider

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/13248492528

## Changes
- add name sanitization and min length validation


twingate-blee (approved):
lgtm

PR #672, opened 2025-04-09
Feature: search groups by name

vmanilo:
Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/14346067365

## Changes
- added search by name for cached resources

### DIFF
--- a/twingate/internal/model/group.go
+++ b/twingate/internal/model/group.go
@@ -1,6 +1,11 @@
 package model
 
-import "github.com/Twingate/terraform-provider-twingate/v3/twingate/internal/attr"
+import (
+	"github.com/Twingate/terraform-provider-twingate/v3/twingate/internal/attr"
+	"regexp"
+	"slices"
+	"strings"
+)
 
 const (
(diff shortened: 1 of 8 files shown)


twingate/internal/model/group.go
```diff
@@ -36,6 +41,67 @@ func (g Group) ToTerraform() interface{} {
...
+		return false
+	}
+
+	// filter by type
+	if len(filter.GetTypes()) > 0 && !slices.Contains(filter.GetTypes(), g.Type) {
+		return false
+	}
+
+	// filter by name
+	if filter.GetName() != "" {
```
twingate-blee, line 67: A nitpick if you change this to: 

`if name := filter.GetName(); name != "" {`

you can change where you have: 

`filter.GetName()` to `name`

example

```
		case attr.FilterByContains:
			if !strings.Contains(g.Name, filter.name()) {
				return false
			}
```

twingate-blee (approved):
lgtm

PR #673, opened 2025-04-11
[Workflows] Ignore issues with the backlog label

alexb-twingate:
A few planned issues were closed as stale. This change adds the `backlog` label as an exempt issue label so that we can mark any issues that are planned.


twingate-blee (approved):
lgtm

PR #675, opened 2025-04-15
Feature: search groups by name

vmanilo:
Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/14461416058

## Changes
- added possibility to search groups in cache by name (with support of contains, suffix, prefix, exclude and regexp) which should help reduce read API calls mainly for datasources


twingate-blee (approved):
lgtm

PR #682, opened 2025-04-23
Feature: search resource in cache by name

vmanilo:
Datasource - read resources improvements

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/14624352908

## Changes
- read resources from cache on datasource usage


twingate/internal/model/resource.go
```diff
@@ -106,12 +108,64 @@ type ResourceFilter interface {
 	GetFilterBy() string
 	GetTypes() []string
 	GetIsActive() *bool
+	GetTags() map[string]string
 }
 
 func (r Resource) Match(filter ResourceFilter) bool {
```
twingate-blee, line 114: @vmanilo This function is too large. Lint throws the error "calculated cyclomatic complexity for function Match is 19, max is 10"

vmanilo: it's fixed now with updated linter settings

twingate-blee (approved):
lgtm

Issue #684, opened 2025-04-24
[Bug] Twingate groups >50 members produce constant plan changes


robotparty:
### Summary

We have 3 Twingate groups with over 50 members and every plan shows changes to membership for all users above 50.

### Steps to reproduce
Create a twingate group with over 50 users as members, run a TF plan or apply

### Actual outcome
For all users over 50 in the group, the plan shows membership changes but if it's applied, nothing actually changes in twingate. 

### Expected outcome
No plan changes unless we actually change membership

### Twingate provider version

v3.0.15

### Terraform version

1.11.4

### Terraform output

```shell
# module.twingate_groups.twingate_group.current["dev-dashboard"] will be updated in-place
!   resource "twingate_group" "current" {
        id                 = "<redacted>"
!       is_authoritative   = true -> (known after apply)
        name               = "dev-dashboard"
!       security_policy_id = "U2VjdXJpdHlQb2xpY3k6NjY1ODA=" -> (known after apply)
!       user_ids           = [
+           "v=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
            # (50 unchanged elements hidden)
        ]
    }

  # module.twingate_groups.twingate_group.current["prod-idp"] will be updated in-place
!   resource "twingate_group" "current" {
        id                 = "<redacted>"
!       is_authoritative   = true -> (known after apply)
        name               = "prod-idp"
!       security_policy_id = "U2VjdXJpdHlQb2xpY3k6NjY1ODA=" -> (known after apply)
!       user_ids           = [
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
+           "<redacted>=",
[...]

PR #686, opened 2025-04-25
Fix: fetch all users for group

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/684

Tests passes[UPD]: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/14671048480

## Changes
- fix fetching all users for groups on cache init


twingate/internal/client/group.go
```diff
@@ -146,6 +156,12 @@ func (client *Client) ReadFullGroups(ctx context.Context) ([]*model.Group, error
 		return nil, err //nolint
 	}
 
+	for i, group := range response.Edges {
+		if err := response.Edges[i].Node.Users.FetchPages(oprCtx, client.readGroupUsersAfter, newVars(pageLimit(client.pageLimit), gqlID(group.Node.ID))); err != nil {
+			return nil, err //nolint
```
twingate-blee, line 161: same comment as above

twingate-blee:
@vmanilo Instead of using the same error message "failed to read users for group" can you make it so it is relevant to the function ie CreateGroup, ReadFullGroups, etc..

twingate-blee (approved):
lgtm

Issue #687, opened 2025-04-25
[Bug] v2 to v3 migration fails with: `Error decoding from prior state: unsupported attribute "access"`


tjstansell:
### Summary

v3 provider cannot read v2 terraform_resource from state file

### Steps to reproduce
I updated our provider from v2.1.2 to v3.0.18 and converted the `access {}` blocks to `access_group {}` and `access_service {}` blocks per the migration guide.

### Actual outcome
plan resulted in warnings like this for every resource we have:
```
│ Warning: Failed to decode resource from state
│
│ Error decoding "twingate_resource.this[\"MYRESOURCE\"]" from prior state: unsupported attribute "access"
```
with a subsequent error:
```
│ Error: Value Conversion Error
│
│   with twingate_resource.this["MYRESOURCE"],
│   on resources.tf line 1, in resource "twingate_resource" "this":
│    1: resource "twingate_resource" "this" {
│
│ An unexpected error was encountered while verifying an attribute value matched its expected type to prevent unexpected behavior or panics. This is always an error in the provider. Please report the following to the provider developer:
│
│ Expected framework type from provider logic: types.MapType[basetypes.StringType] / underlying type: tftypes.Map[tftypes.String]
│ Received framework type from provider logic: types.MapType[!!! MISSING TYPE !!!] / underlying type: tftypes.Map[tftypes.DynamicPseudoType]
│ Path: tags
```

### Expected outcome
a successful plan with no changes.

### Twingate provider version

v3.0.18

### Terraform version

v1.11.3

### Terraform output

```shell

```

### Additional information

Given that the fatal errors were referring a a `tags` attribute that was introduced in v3.0.17, I downgraded our provider to v3.0.16 and ran another terraform plan.  This appears to have "succeeded", but produced the following final warning:
```
No changes. Your infrastructure matches the configuration.

Terraform has compared your real infrastructure against your configuration and found no differences, so no changes are needed.
╷
│ Warning: Please update the access blocks.
│
│   with twingate_resource.this["MYRESOURCE"],
│   on resources.tf line 1, in resource "twingate_resource" "this":
│    1: resource "twingate_resource" "this" {
│
│ See the v2 to v3 migration guide in the Twingate Terraform Provider documentation https://registry.terraform.io/providers/Twingate/twingate/latest/docs/guides/migration-v2-to-v3-guide
│
│ (and 166 more similar warnings elsewhere)
╵
```

[...]

PR #691, opened 2025-04-28
Fix missing null value for tags on resource upgrade

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/687

Tests passed: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/14718366033

## Changes
- added missing `tags` attribute with null value


twingate-blee (approved):
lgtm

Issue #688, opened 2025-04-26
[Bug]


tjstansell:
### Summary

documentation for importing a twingate_dns_filtering_profile resource

### Steps to reproduce
None
### Twingate provider version

v3.0.18

### Terraform version

any

### Terraform output

```shell

```

### Additional information

After playing with dns filtering via the console, I went to add it to our terraform and noticed the [documentation](https://registry.terraform.io/providers/Twingate/twingate/latest/docs/resources/dns_filtering_profile) for the `twingate_dns_filtering_profile` resource doesn't mention support for importing an existing profile.  I was able to import, however, by passing the profile id, so it does seem to be supported.  It should be documented.

PR #706, opened 2025-05-13
Fix: update docs with import for dns_filtering_profile

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/688

## Changes
- updated docs with import example for dns_filtering_profile


twingate-blee (approved):
Lgtm

Issue #716, opened 2025-05-19
[Bug] Alias is not considered on import


rgravlin-figure:
### Summary

Importing a resources ignores the associated alias field

### Steps to reproduce
1. Create a resource via the UI that uses a Standard Address and include a DNS alias (https://www.twingate.com/docs/resource-aliases)
2. Import the resource via Terraform and set the `alias` field to match that of the resource that was created
3. You will see it wants to `~` change the resource upon import

### Actual outcome
The provider does not seem to have knowledge of the resource's alias upon import

It is unknown whether or not the provider strips the resource of the `alias` field when the import is completed.  Since there are no changes expected I don't believe that it will.

The issue will be if things needed to be re-created they would not be correct.

### Expected outcome
The plan should produce a diff if the `alias` is missing from the imported resource and it was created with one, or should produce no diff if the proper `alias` is applied to the same resource.

### Twingate provider version

3.2.0

### Terraform version

1.11.4

### Terraform output

```shell
# twingate_resource.this["sub.example.com"] will be updated in-place
# (imported from "UmVzb3VyY2U6MTMzNw==")
~ resource "twingate_resource" "this" {
      address                     = "1.2.3.4"
    + alias                       = "sub.example.com"
      approval_mode               = "MANUAL"
      id                          = "UmVzb3VyY2U6MTMzNw=="
      is_active                   = true
      is_authoritative            = false
      is_browser_shortcut_enabled = false
      is_visible                  = true
      name                        = "sub.example.com"
      protocols                   = {
```

PR #717, opened 2025-05-19
Fix: import resource with alias attribute

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/716

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/15121061324

## Changes
- added missing `alias` attribute on resource import


twingate-blee (approved):
lgtm

PR #729, opened 2025-06-06
Resource upgrade test v1

vmanilo:
Resolves issues with upgrade resource from v1 to v2

## Changes
- added tests for upgrader v1
- fixed issues

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,7 @@ on:
       - 'README.md'
     branches:
       - main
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -16,6 +16,7 @@ on:
       - 'README.md'
     branches:
       - main
+      - resource-upgrade-test-v1
```
twingate-blee, line 19: remove this

PR #737, opened 2025-06-23
Resource upgrade test v0

vmanilo:
Resolves issues with upgrading resource from v0 to v1

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/15832851276

## Changes
- added tests to cover resource upgrade phase for schema migration from v0 to v1


twingate-blee (approved):
lgtm

PR #748, opened 2025-07-29
feat: expose cache configuration to environment variables

zhuoyang:
## Changes
- expose cache configuration to `TWINGATE_CACHE_GROUPS` and `TWINGATE_CACHE_RESOURCES` env var

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -60,7 +60,9 @@ This is optional and shouldn't be changed under normal circumstances.
 Optional:
 
 - `groups_enabled` (Boolean) Specifies whether the provider should cache groups. The default value is `true`.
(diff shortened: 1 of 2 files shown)


zhuoyang:
@vmanilo that would be helpful

@twingate-blee we will still need to add in the provider configuration for all of our terraform workspace, which is a lot of work, this PR allow us to set the cache configuration on the runner level.

zhuoyang:
@twingate-blee any reason not to implement this ? Imho it is pretty harmless to have this features exposed as environment variables.

twingate-blee:
@zhuoyang we are currently working on pre-populate cache based on tags and it might replace this feature

Issue #757, opened 2025-08-20
Provider initialization can take tens of minutes due to throttles


uristernik:
### Summary

Provider initialization can take tens of minutes

### What happened?

Hello

We provision twingate resource using Terraform. 
During our version deployment we are running terraform plan on roughly ~120 workspaces.

We noticed that initializing the Twingate terraform provider, without calling any resources / data sources can take >10 minutes due to throttled calls. I am emphasizing that this is only to initialize the provider not calling any resources / data sources.

Here is a log line from one of our Terraform runs:

```
{
  "@caller": "github.com/hashicorp/terraform-plugin-go@v0.26.0/tfprotov6/internal/tf6serverlogging/downstream_request.go:42",
  "@level": "trace",
  "@message": "Received downstream response",
  "@module": "sdk.proto",
  "@timestamp": "2025-08-19T18:53:14.501411Z",
  "diagnostic_error_count": 0,
  "diagnostic_warning_count": 0,
  "tf_proto_version": "6.8",
  "tf_provider_addr": "registry.terraform.io/Twingate/twingate",
  "tf_req_duration_ms": 568337,
  "tf_req_id": "5f314651-8a9b-23c2-f5e3-af1753ca42b8",
  "tf_rpc": "ConfigureProvider",
  "timestamp": "2025-08-19T18:53:14.501Z"
}
```

Why configuring the provider without any resources / data sources are being throttled? I assume it's related to https://github.com/Twingate/terraform-provider-twingate/issues/731? Should we disable cache? Our current provider is configured with default values

```
provider "twingate" {
  api_token = var.twingate_api_key
  network   = local.config["twingate_org_name"]
}
```

### Steps to reproduce

Big tenant with many workspaces running at the same time

### Actual outcome

Terraform plan time increases to >10 minutes

### Expected outcome

Quick plan

### Twingate provider version

v3.0.18

### Terraform version

v1.12.0

### Terraform output

```shell

```

uristernik:
Consider caching the groups into `.terraform` so it will persist between terraform runs?

uristernik:
I think I got an answer to my question, it does fetch all group when using cache

https://github.com/Twingate/terraform-provider-twingate/blob/3ce0a34ada9b2c8f0e8efe07b9c0a3164661756c/twingate/internal/client/cache.go#L28-L31
[...]

PR #763, opened 2025-09-04
Cache filter config

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/757

Tests pases: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/17665003063

## Changes
- added cache config


.github/workflows/ci.yml
```diff
@@ -16,6 +16,7 @@ on:
       - 'README.md'
     branches:
       - main
+      - cache-filter-config
```
twingate-blee, line 19: This can be removed now

Issue #760, opened 2025-08-26
[Feature] Add role update support for IDP users


bertekintw:
### Summary

Add role update support for IDP users

### Why this is useful

Today we only seem to support role changes for non-IDP users via Terraform Provider, while this is possible to do via Admin Console. 
https://github.com/Twingate/terraform-provider-twingate/blob/main/twingate/internal/provider/resource/user.go#L213

For feature parity, we should add role update support for IDP users too.

### How it should work

Role Update for IDP users should work via Terraform Provider too, similar to non-IDP users.

PR #770, opened 2025-09-13
Feature: allow user role update for IDP users

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/760

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/17692050564

## Changes
- allow user role update for IDP users


twingate-blee (approved):
lgtm

PR #797, opened 2025-11-23
Fix resource import security_policy_id

vmanilo:
Resolves issue with missing security_policy_id on resource import

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/19611996609

## Changes
- added security_policy_id on resource import


twingate-blee (approved):
lgtm

PR #862, opened 2026-04-12
Fix resource alias

### LINKED ISSUE #861: [Bug] alias attribute causes "Provider produced invalid plan" when set to null
### Summary

When alias is set to null on a twingate_resource (i.e. the resource has no alias), the provider's PlanResourceChange normalizes null to "" internally. Terraform then detects a mismatch between the config value (null) and the planned value (""), and aborts with an error

### What happened?

Since the original bug report is already closed, I am opening a new one:

The closed issue is here: https://github.com/Twingate/terraform-provider-twingate/issues/850

### Twingate provider version

v4.1.0

### Terraform version

OpenTofu v1.11.5

### Terraform output

```shell

```

### Additional information

_No response_

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/861

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/24313399066

## Changes
- fix resource alias update behaviour


twingate-blee (approved):
lgtm

PR #864, opened 2026-04-15
Fix resource tags

vmanilo:
Resolves failing smoke tests

Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/24456050217

## Changes
- fix for resource import
- fix for handling default tags


twingate-blee (approved):
lgtm

PR #885, opened 2026-06-23
Added support for multiverse

vmanilo:
Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/28033609064

## Changes
- added support for multiverse by resolving regional host from redirect request


twingate-blee (approved):
lgtm

PR #889, opened 2026-07-07
Feature: added routing_mode to Resource

### LINKED ISSUE #888: Add resource exclusion support (routing_mode) to twingate_resource
**New functionality**
- Update `twingate_resource` resource
  - Optional
    - `routing_mode` (String): Controls whether traffic to this resource is routed through Twingate or bypassed. Valid values are `THROUGH_TWINGATE` (default) and `BYPASS_TWINGATE`.

**Behavior**
- We're adding the ability to mark a resource as a bypass resource — traffic to that resource is excluded from Twingate routing. This allows specific addresses to be excluded from being captured by wildcard resources.
- `THROUGH_TWINGATE` (default): Traffic is routed normally through Twingate.
- `BYPASS_TWINGATE`: Traffic to this resource bypasses Twingate entirely.
- Constraints enforced on `BYPASS_TWINGATE` resources:
  - Must be of network type
  - Cannot have a security policy set (on the resource or its access groups)
  - Cannot be a wildcard address
- Existing resources will be treated and set as `THROUGH_TWINGATE` but we should not show a change in Terraform state since it is the existing behavior.

**Example usage**
```terraform
resource "twingate_resource" "bypass_example" {
  name              = "Internal Bypass Resource"
  address           = "public.example.com"
  remote_network_id = data.twingate_remote_network.example.id
  routing_mode      = "BYPASS_TWINGATE"
}

resource "twingate_resource" "default_example" {
  name              = "Normal Resource"
  address           = "*.example.com"
  remote_network_id = data.twingate_remote_network.example.id
  # routing_mode defaults to THROUGH_TWINGATE
}


```

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/888

Test passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/28865100433

## Changes
- added new attribute `routing_mode` to Resource


twingate-blee (approved):
lgtm

PR #905, opened 2026-08-14
feat: Support twingate web app resource type 

### LINKED ISSUE #904: Add support for Web App resource
**New functionality**
- Create new `twingate_web_app_resource` resource type

**Behavior**
- Add ability to create, update and delete `WEB_APP` resource type.
  - Add `downstream.port` and `upstream.port` attributes, which must be between 1 and 65535 inclusive
  - Add `request_header_rewrites` attribute which consists key-value pairs that correspond with header to be rewritten to the upstream resource.

**Example usage**
```terraform
resource "twingate_remote_network" "prod" {
  name = "Production Network"
}

resource "twingate_x509_certificate_authority" "tls" {
  name        = "My TLS CA"
  certificate = file("ca.pem")
}

resource "twingate_gateway" "main" {
  remote_network_id = twingate_remote_network.prod.id
  address           = "10.0.0.1:8443"
  x509_ca_id        = twingate_x509_certificate_authority.tls.id
}

# Web App resource
resource "twingate_web_app_resource" "Grafana" {
  name              = "Grafana"
  address           = "grafana.int"
  remote_network_id = twingate_remote_network.prod.id
  # Other existing fields supported on `twingate_resource` except `protocols`
  # ...

  # New attributes
  downstream = {
    port = 80
  }
  upstream = {
    port = 8080
  }
  request_header_rewrites = {
    x-twingate-user = "{{user}}"
    x-twingate-groups = "{{groups}}"
  }
}
```

clement0010:
Resolves #904

## Changes

Create new `twingate_web_app_resource` resource type
- Most attributes are the same as  `twingate_kubernetes_resource` except `protocols` attribute is omitted
- Support new `downstream`, `upstream` and `request_header_rewrites` attributes

Example: 
```terraform
resource "twingate_remote_network" "prod" {
  name = "Production Network"
}

resource "twingate_x509_certificate_authority" "tls" {
  name        = "My TLS CA"
  certificate = file("ca.pem")
}

resource "twingate_gateway" "main" {
  remote_network_id = twingate_remote_network.prod.id
  address           = "10.0.0.1:8443"
  x509_ca_id        = twingate_x509_certificate_authority.tls.id
}

# Web App resource
resource "twingate_web_app_resource" "Grafana" {
  name              = "Grafana"
  address           = "grafana.int"
  remote_network_id = twingate_remote_network.prod.id
  # Other existing fields supported on `twingate_resource` except `protocols`
  # ...

  # New attributes
  downstream = {
    port = 80
  }
  upstream = {
    port = 8080
  }
  request_header_rewrites = {
    x-twingate-user = "{{username}}"
    x-twingate-groups = "{{groups}}"
  }
}
```

## Notes

- **No `protocols` attribute.** The API derives a web app resource's TCP/UDP ports from `downstream.port` instead.


twingate/internal/customvalidator/mapvalidator_case_insensitive_unique_keys.go
clement0010, line 1: I decided not to reuse the `MapValidator` for `tags` because it is case-insensitive, e.g. `Foo: Bar` and `foo: bar` are two different tags. In the context of HTTP header, they should be the same.

twingate/internal/test/acctests/resource/web-app-resource_test.go
```diff
@@ -0,0 +1,330 @@
...
+
+func TestAccTwingateWebAppResourceDeleteNonExisting(t *testing.T) {
+	t.Parallel()
+
+	setup := newWebAppTestSetup(t, "10.0.0.6:8080")
+
+	sdk.Test(t, sdk.TestCase{
+		ProtoV6ProviderFactories: acctests.ProviderFactories,
+		PreCheck:                 func() { acctests.PreCheck(t) },
+		TerraformVersionChecks:   acctests.VersionCheckForWriteOnlyAttributes(),
```
vmanilo, line 359: I think this check can be removed, we used it mainly for x509CertificateAuthority resource which has write-only field

```
attr.Certificate: schema.StringAttribute{
				Required:    true,
				WriteOnly:   true,
				Description: "The PEM-encoded X509 certificate. This field is write-only and will not be returned by the API.",
				Validators: []validator.String{
					customvalidator.Certificate(),
				},
			}
```

twingate/internal/test/acctests/resource/web-app-resource_test.go
```diff
@@ -0,0 +1,330 @@
...
+		resourceName:        test.RandomName(),
+		resourceAddress:     "internal.acme.com",
+	}
+}
+
+func (s webAppTestSetup) config(upstreamPort, downstreamPort int, extra string) string {
+	return s.prereqs + terraformResourceWebApp(s.webAppTFName, s.gatewayTFName, s.remoteNetworkTFName,
+		s.resourceName, s.resourceAddress, upstreamPort, downstreamPort, extra)
+}
+
```
vmanilo, line 87: I think we also need to add acctest for resource import, often some fields doesn't import out of the box automatically

twingate/internal/client/web-app-resource.go
```diff
@@ -0,0 +1,195 @@
...
+
+	response := query.CreateWebAppResource{}
+
+	if err := client.mutate(ctx, &response, variables, opr, attr{name: webAppResource.Name}); err != nil {
+		return nil, err
+	}
+
+	res := response.ToModel()
+	if res == nil {
+		return nil, nil //nolint:nilnil
```
vmanilo: lets check if this case does make sense

twingate/internal/client/web-app-resource.go
```diff
@@ -0,0 +1,195 @@
...
+
+	response := query.UpdateWebAppResource{}
+
+	if err := client.mutate(ctx, &response, variables, opr, attr{id: webAppResource.ID}); err != nil {
+		return nil, err
+	}
+
+	res := response.ToModel()
+	if res == nil {
+		return nil, nil //nolint:nilnil
```
vmanilo: this case looks odd to me

twingate/internal/client/web-app-resource.go
```diff
@@ -0,0 +1,195 @@
...
+		},
+		func(edge *query.ShallowResourceEdge) *model.WebAppResource {
+			return &model.WebAppResource{
+				ID:   string(edge.Node.ID),
+				Name: edge.Node.Name,
+			}
+		}), nil
+}
+
+func (client *Client) DeleteWebAppResource(ctx context.Context, resourceID string) error {
```
vmanilo, line 181: I think we could use existing function to delete resource, this maybe useful if we want to have exact logs that user trigger removal of web-app-resource

twingate/internal/client/web-app-resource_test.go
```diff
@@ -0,0 +1,73 @@
...
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			client := newTestClient(t.Context())
+			httpmock.ActivateNonDefault(client.HTTPClient)
+			defer httpmock.DeactivateAndReset()
+
+			httpmock.RegisterResponder("POST", client.GraphqlServerURL,
+				httpmock.NewStringResponder(200, c.responseBody))
+
+			resources, err := client.ReadWebAppResources(context.Background())
```
vmanilo, line 62: would be better to cover with tests other new methods as well

twingate-blee:
The `twingate_web_app_resource` resource is not inheriting the provider tags.

Example

```provider "twingate" {
,,,

  default_tags = {
    tags = {
      managed-by = "Terraform"
      owner      = "DevOps"
      cloud      = "GCP"
    }
  }
}

The other resources picks up the tags
<img width="1234" height="377" alt="Screenshot 2026-08-14 at 5 06 18 PM" src="https://github.com/user-attachments/assets/244f720c-f51b-479c-92f2-d5ec15ba1ea9" />

File: docs/resources/web_app_resource.md [hunk 7]
```diff
@@ -0,0 +1,123 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "twingate_web_app_resource Resource - terraform-provider-twingate"
+subcategory: ""
+description: |-
+  Web App Resources are Twingate resources accessed via a Gateway, which proxies HTTP traffic from a downstream port to an upstream port.
```
minhtule: > ... which proxies HTTP traffic from a downstream port to an upstream port.

I think we should drop this part because "downstream port" in this text sounds like the port that the Gateway listens to.

docs/resources/web_app_resource.md
```diff
@@ -0,0 +1,123 @@
...
+---
+
+# twingate_web_app_resource (Resource)
+
+Web App Resources are Twingate resources accessed via a Gateway, which proxies HTTP traffic from a downstream port to an upstream port.
```
minhtule: Ditto

docs/resources/web_app_resource.md
```diff
@@ -0,0 +1,123 @@
...
+}
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `address` (String) The address of the Web App Resource (IP or FQDN).
+- `downstream` (Attributes) The downstream configuration. The connection between the protocol client and the Gateway. (see [below for nested schema](#nestedatt--downstream))
```
minhtule, line 60: I would group `downstream` and `upstream` together, or is this list alphabetically sorted?

twingate/internal/model/web-app-resource.go
```diff
@@ -0,0 +1,26 @@
...
+	Name                  string
+	Address               string
+	GatewayID             string
+	RemoteNetworkID       string
+	IsVisible             *bool
+	Alias                 *string
+	SecurityPolicyID      *string
+	Tags                  map[string]string
+	UpstreamPort          int64
+	DownstreamPort        int64
```
minhtule: Should we model upstream and downstream as objects (or structs) so we could easily add `tls` for HTTPS support?

twingate/internal/customvalidator/mapvalidator_case_insensitive_unique_keys_test.go
```diff
@@ -0,0 +1,69 @@
...
+		{
+			name:        "keys differing only by case - error",
+			input:       buildMap("X-Twingate-User", "x-twingate-user"),
+			expectError: true,
+		},
+		{
+			name:        "mixed case collision among distinct keys - error",
+			input:       buildMap("x-a", "X-A", "x-b"),
+			expectError: true,
+		},
```
minhtule, line 50: not sure how these 2 tests are different? the first one is a sub-set of the 2nd one? If I change the `input` in the 1st test to `buildMap("X-Twingate-User", "x-twingate-user", "foo")`, the test would be equivalent to the 2nd test?

twingate/internal/provider/resource/web-app-resource.go
```diff
@@ -0,0 +1,436 @@
...
+}
+
+func (r *webAppResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	resource.ImportStatePassthroughID(ctx, path.Root(attr.ID), req, resp)
+}
+
+//nolint:funlen
+func (r *webAppResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		Description: "Web App Resources are Twingate resources accessed via a Gateway, which proxies HTTP traffic from a downstream port to an upstream port.",
```
minhtule: Ditto. Should drop ", which proxies HTTP traffic from a downstream port to an upstream port."

twingate/internal/provider/resource/web-app-resource.go
```diff
@@ -0,0 +1,436 @@
...
+
+func webAppPortObject(ctx context.Context, port int64) (types.Object, diag.Diagnostics) {
+	return types.ObjectValueFrom(ctx, webAppPortAttributeTypes, webAppPortModel{Port: types.Int64Value(port)})
+}
+
+// getHeaderRewrites converts the configured map for the API. Header rewrites and
+// tags are both plain string maps, so the tag converter applies: a null, unknown
+// or empty map becomes nil, which the client sends as an empty list.
+func getHeaderRewrites(rawRewrites types.Map) map[string]string {
+	return getTags(rawRewrites)
```
minhtule, line 234: nit: would be nicer to rename `getTaps` to smth like `getKeyValueMap` so the usage is clearer and we don't need this long comment. Probably better to do it in another PR.

clement0010:
> The `twingate_web_app_resource` resource is not inheriting the provider tags.
> 
> Example
> 
> ```
> 
>   default_tags = {
>     tags = {
>       managed-by = "Terraform"
>       owner      = "DevOps"
>       cloud      = "GCP"
>     }
>   }
> }
> ```
> 
> The other resources picks up the tags
> Screenshot 2026-08-14 at 5 06 18 PM
> ```

@twingate-blee It seems like `default_tags` is not supported for K8s and SSH resource type as well. I think it'd be nicer to add `default_tags` support for all other PAM resource types in a separate PR.

twingate/internal/test/acctests/resource/web-app-resource_test.go
clement0010: We need it because the web app resource needs the x509 CA, this test has the same setup as K8s resource

docs/resources/web_app_resource.md
clement0010: It is generated by the `tfplugindocs`

twingate/internal/client/web-app-resource.go
clement0010: You're right, both branches are dead. `client.mutate` already returns `ErrGraphqlResultIsEmpty` when `resp.IsEmpty()` is true, and `IsEmpty()` and `ToModel()` test the same `Entity == nil`, so by the time we get the result `Entity` is non-nil.

I removed it in [46eb449](https://github.com/Twingate/terraform-provider-twingate/pull/905/commits/46eb4492ec1a9568d537b3b15461a9bce89de1ad)

clement0010: We have a similar issue with [`kubernetes_resource`](https://github.com/Twingate/terraform-provider-twingate/blob/3829d276e50cde8bc87e59ddeced9db78f119aa8/twingate/internal/client/kubernetes-resource.go#L37-L39) and [`ssh_resource`](https://github.com/Twingate/terraform-provider-twingate/blob/3829d276e50cde8bc87e59ddeced9db78f119aa8/twingate/internal/client/ssh-resource.go#L37-L39), we should probably clean it up in a separate PR

twingate/internal/client/web-app-resource.go
clement0010: Fixed in [46eb449](https://github.com/Twingate/terraform-provider-twingate/pull/905/commits/46eb4492ec1a9568d537b3b15461a9bce89de1ad)

twingate/internal/client/web-app-resource.go
clement0010: We have dedicated functions for `DeleteSSHResource` and `DeleteKubernetesResource` too. While they issue the same `query.DeleteResource` mutation, the only difference is the operation label, so errors and logs read "Web App resource" instead of the generic "resource". If we want to combine them, I'd prefer we keep this in this PR and refactor all three function in a separate PR

twingate/internal/test/acctests/resource/web-app-resource_test.go
clement0010: Added in [0fb2462](https://github.com/Twingate/terraform-provider-twingate/pull/905/commits/0fb2462db822650a82b8b46109eea7482991647c)

twingate/internal/client/web-app-resource_test.go
clement0010: Good call, I added in [dba50b0](https://github.com/Twingate/terraform-provider-twingate/pull/905/commits/dba50b0ac1b81e4048577e09342bbe3e590e7767)

docs/resources/web_app_resource.md
clement0010: Dropped in [c52973a](https://github.com/Twingate/terraform-provider-twingate/pull/905/commits/c52973a07a0a4dd35bcf01340409e0c3fb8b424c)

twingate/internal/customvalidator/mapvalidator_case_insensitive_unique_keys_test.go
clement0010: You're right, they are similar, i've combined them in [609c00a](https://github.com/Twingate/terraform-provider-twingate/pull/905/commits/609c00a6817963875218a8db57babe416c6b948f)

twingate/internal/model/web-app-resource.go
clement0010: Yeap, I think it's easier to extend if they are a `struct`. Fixed in [7ea504a](https://github.com/Twingate/terraform-provider-twingate/pull/905/commits/7ea504a4b062650debdea9a3c55855d8dd263899)

twingate-blee (approved):
lgtm

minhtule (approved):
LGTM 👍

twingate/internal/client/web-app-resource.go
```diff
@@ -0,0 +1,191 @@
...
+		cursor(query.CursorResources),
+		pageLimit(client.pageLimit),
+	)
+
+	response := query.ReadShallowResourcesWithType{}
+	if err := client.query(ctx, &response, variables, opr, attr{id: "All"}); err != nil && !errors.Is(err, ErrGraphqlResultIsEmpty) {
+		return nil, err
+	}
+
+	if err := response.FetchPages(ctx, client.readShallowResourcesWithTypeAfter, variables); err != nil {
```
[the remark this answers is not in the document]
vmanilo: I think we can ignore this copilot comment, but we can pass down operation ctx like this, to keep reference to original call

```suggestion
	if err := response.FetchPages(withOperationCtx(ctx, opr), client.readShallowResourcesWithTypeAfter, variables); err != nil {
```

twingate/internal/test/acctests/resource/web-app-resource_test.go
vmanilo: ok, I see, make sense

vmanilo (approved):
looks good to me

PR #906, opened 2026-08-14
Require IMDSv2 in the AWS EC2 deployment guide example

neth-tg:
Resolves [DEV-1366](https://linear.app/twingate/issue/DEV-1366/update-tf-provider-documentation-to-disable-imdsv1)

## Changes

The AWS EC2 deployment guide's example launched the Connector with the `terraform-aws-modules/ec2-instance` module defaults, which set `http_tokens = "optional"` — leaving IMDSv1 enabled. Most security teams and scanners flag that on EC2, so the example now requires IMDSv2 explicitly:

```terraform
metadata_options = {
  http_endpoint               = "enabled"
  http_tokens                 = "required"
  http_put_response_hop_limit = 1
}
```

This is a documentation and example change only. No provider behavior is affected.

Note that the module's `metadata_options` variable is a `map(string)` that replaces its default map rather than merging with it, so `http_put_response_hop_limit` is set explicitly to preserve the previous value.

### Also in this change

- Fixed a dangling reference in the "Creating Resources" example: `twingate_remote_network.my_aws_network.id` pointed at a network the guide never defines. It now references `aws_network`, matching the earlier block.
- Ran `terraform fmt` on the guide's inline snippets.

## Verification

- `terraform fmt -check` passes on `examples/ami/ami.tf` and every `terraform` snippet in the guide.
- `docs/guides/aws-deployment-guide.md` was verified to match the rendered `templates/guides/aws-deployment-guide.md.tmpl`, so `make ci-checks` should find no drift.
- Not applied against a live AWS account — the change is confined to documentation and the example file.

### DIFF
--- a/docs/guides/aws-deployment-guide.md
+++ b/docs/guides/aws-deployment-guide.md
@@ -61,16 +61,9 @@ data "aws_ami" "latest" {
       "twingate/images/hvm-ssd/twingate-amd64-*",
     ]
   }
-    # Disable IMDSv1
-    metadata_options {
-    http_tokens            = "required"
-    http_endpoint          = "enabled"
-  }
   owners = ["617935088040"]
 }
 ```
-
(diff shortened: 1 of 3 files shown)


vmanilo (approved):

twingate-blee (approved):
lgtm

PR #913, opened 2026-08-20
feat: Remove `protocols` attribute from `twingate_ssh_resource` and `twingate_kubernetes_resource`

### LINKED ISSUE #911: Remove `protocols` attribute for K8s and SSH resource
### New functionality

Remove `protocols` attribute from `twingate_kubernetes_resource` and `twingate_ssh_resource` resource types

### Notes
This is a breaking change, we will include this change in v5 release

clement0010:
Resolves #911

## Changes

Port restrictions are not used for SSH and Kubernetes resource. Therefore the  `protocols` attribute on `twingate_ssh_resource` and `twingate_kubernetes_resource` does not do anything meaningful, so it should be removed from the Terraform schema, the models, the create and update mutations, and the read queries.

`twingate_resource` is untouched and keeps `protocols`, so the shared helpers (`protocols()`, `convertProtocols`, `newProtocolsInput`, `attr.Protocols`) all stay in place.

## Notes

⚠️ Breaking change, targeted at the v5 release. A configuration that still sets `protocols` on either resource fails with `Unsupported argument`.

The v4 → v5 migration guide is deliberately left out of this PR and will follow separately.


minhtule (approved):
LGTM 👍

twingate-blee (approved):
lgtm

vmanilo:
lets discard last commit, it's not necessary and code will be simpler

PR #902, opened 2026-08-12
Remove `twingate_gateway_config`

clement0010:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/903

## Changes

The Gateway no longer reads upstreams from its config file (Twingate/gateway#286). SSH upstream address and username come from the runtime connection, and in-cluster Kubernetes is implicit. With the resource lists gone, `twingate_gateway_config` only ever renders a static YAML document, and it needs nothing from Terraform state to do it, so `templatefile()` covers the same ground with less machinery.

### Breaking change

`twingate_gateway_config` shipped in v4.1.0 and is now removed. Render the config with `templatefile()` as the SSH gateway guides show, then drop the resource from state:

```
terraform state rm twingate_gateway_config.<name>
```

It was purely local and never created a remote object, so nothing else is required.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #916, opened 2026-08-21
Remove `username` from `twingate_ssh_resource`

### LINKED ISSUE #903: Remove `twingate_gateway_config` support
The Gateway removed `upstreams` from its configuration (Twingate/gateway#286). SSH upstream address and username come from the runtime connection, and in-cluster Kubernetes is implicit. With the resource lists gone, `twingate_gateway_config` only ever renders a static YAML document, and it needs nothing from Terraform state to do it, so `templatefile()` covers the same ground with less machinery.

## Desired outcome
- Remove `twingate_gateway_config` support

clement0010:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/903

Branched from #902.

## Changes

`username` never reached the API and its only consumer was `twingate_gateway_config`, which rendered it into `ssh.upstreams[].user`. Now that #902 removes that resource, the attribute is a silent no-op.

### Breaking change

Configs setting `username` on `twingate_ssh_resource` fail with `Unsupported argument`.


minhtule (approved):
LGTM 👍

twingate-blee (approved):
lgtm

PR #922, opened 2026-08-25
feat: Add `tls` attribute to web app resource `upstream` and `downstream`

### LINKED ISSUE #921: Add `tls` attribute to web app resource type to support HTTPS
**New functionality**
- Add `tls` attribute to `downstream` and `upstream` attribute to support HTTPS

**Behavior**
- Set `tls` boolean flag to `true/false` for the web app resource

**Example usage**
```terraform
# Web App resource
resource "twingate_web_app_resource" "Grafana" {
  # New attributes
  downstream = {
    port = 443
    # New attribute
    tls = true 
  }
  upstream = {
    port = 8443
    # New attribute
    tls = true
  }
}
```

clement0010:
Resolves #921

⚠️ Branched off from #905

## Changes
- Adds a `tls` boolean to the `upstream` and `downstream` blocks of `twingate_web_app_resource`, threaded through the model, the GraphQL input/read structs and the schema. It is Optional + Computed with a `false` default, matching the API (`tls` defaults to false on create, and is non-null on read).

🚧 **Do not merge before the feature is released on the Controller.**


twingate/internal/attr/common.go
```diff
@@ -8,10 +8,9 @@ const (
 	Type              = "type"
 	IsActive          = "is_active"
 	Port              = "port"
-
-	FilterByRegexp   = "_regexp"
-	FilterByContains = "_contains"
-	FilterByExclude  = "_exclude"
-	FilterByPrefix   = "_prefix"
-	FilterBySuffix   = "_suffix"
```
minhtule, line 16: Why this change? Somebody might want to group the `_xyz` constants together.

minhtule (approved):
LGTM 👍

twingate/internal/attr/common.go
clement0010: Reverted, changed this by accident because I added TLS attribute here and later moved to [`web-app-resource.go`](https://github.com/Twingate/terraform-provider-twingate/blob/3471164be07b6415f38e63fff1177374c5f0c953/twingate/internal/attr/web-app-resource.go#L7)

PR #928, opened 2026-08-26
Fix: added proper error when user set non-default address with in_cluster=true

### LINKED ISSUE #927: [Bug] Error: Provider produced inconsistent result after apply for twingate_kubernetes_resource address
### Summary

Error: Provider produced inconsistent result after apply for twingate_kubernetes_resource address

### Steps to reproduce
Create this resource 
```
resource "twingate_kubernetes_resource" "external_cluster" {
  name              = "External K8s"
  address           = "k8s-api.example.com"
  remote_network_id = twingate_remote_network.prod.id
}
```
### Actual outcome
```
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to twingate_kubernetes_resource.external_cluster, provider "provider[\"registry.terraform.io/twingate/twingate\"]" produced an unexpected new value: .address: was
│ cty.StringVal("k8s-api.example.com"), but now cty.StringVal("kubernetes.default.svc.cluster.local").
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.
```
### Expected outcome
No error

### Twingate provider version

v4.3.2 happens on multiple versions

### Terraform version

Terraform v1.14.2

### Terraform output

```shell

```

### Additional information

From slack discussion
I think this is because we set the address to kubernetes.default.svc.cluster.local when inCluster[ attribute](https://registry.terraform.io/providers/Twingate/twingate/latest/docs/resources/kubernetes_resource#in_cluster-1) is unspecified, which default to True. See [code](https://github.com/Twingate/terraform-provider-twingate/blob/f37936adef42152a3377ed42ac55421ae2fa9c51/twingate/internal/provider/resource/kubernetes-resource.go#L221-L223)

adding in_cluster=false  works and we need to add ca_file and bearer_token_file too

i think we need a better error message that says we can't modify the address when in_cluster=true

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/927

Tests passes [UPD]: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/34511807485

## Changes
- added proper error when user set non-default address with in_cluster=true for twingate_kubernetes_resource


docs/resources/kubernetes_resource.md
```diff
@@ -62,10 +62,10 @@ resource "twingate_kubernetes_resource" "external_cluster" {
 
 - `access_group` (Block Set) Restrict access to certain group (see [below for nested schema](#nestedblock--access_group))
 - `access_policy` (Block Set) Restrict access according to JIT access policy (see [below for nested schema](#nestedblock--access_policy))
-- `address` (String) The address of the Kubernetes Resource (IP or FQDN).
+- `address` (String) The address of the Kubernetes Resource (IP or FQDN). It should be omitted when `in_cluster` is set to `true`. If you want to set custom address, `in_cluster` should be set to `false`.
```
minhtule: > If you want to set custom address, `in_cluster` should be set to `false`.

nit: This sentence should a bit off because you have to set the address when deploying the gateway outside of the cluster, not the other way around. I would just omit this sentence.

go.mod
```diff
@@ -17,7 +17,6 @@ require (
 	github.com/stretchr/testify v1.12.1
 	golang.org/x/crypto v0.55.0
 	golang.org/x/sync v0.22.0
-	gopkg.in/yaml.v3 v3.0.1
```
minhtule: Why is this remove in this PR?

vmanilo: I'm usually run `go mod tidy` and it removes unused libs

docs/resources/kubernetes_resource.md
vmanilo: updated

twingate-blee (approved):
lgtm

PR #932, opened 2026-08-31
Add tools/upgrader Go module to dependabot gomod updates

bertekintw:
## Changes

`tools/upgrader` has its own `go.mod`, separate from the root module, but it wasn't covered by Dependabot's `gomod` config, so its dependencies never received automated update PRs.

This adds it to the existing `gomod` entry using the `directories` (plural) field, which accepts a list of paths.


minhtule (approved):
LGTM 👍

minhtule (approved):
LGTM! 👍

PR #958, opened 2026-09-08
[Docs] Regenerate docs after the terraform-plugin-docs bump

ekampf:
## Summary

`Test docs up-to-date` currently fails on every open PR, including on a branch with no changes of its own. The bump to terraform-plugin-docs 0.25.0 in #951 changed the import-section template, but the generated docs were not regenerated in that PR.

This is the output of `make docs` on an otherwise-clean `main` — one added line per resource that documents an import:

```
The [`terraform import` command](https://developer.hashicorp.com/terraform/cli/commands/import) can be used, for example:
```

No provider or schema changes; only generated files.

## Verification

Generated with terraform 1.14.9 (per `.tool-versions`) — an older terraform produces extra spurious changes, notably deleting `docs/ephemeral-resources/connector_tokens.md` and dropping the write-only note from `docs/resources/x509_certificate_authority.md`, so those are not in this diff. After this lands, `make docs` on main is a no-op.

### DIFF
--- a/docs/resources/connector.md
+++ b/docs/resources/connector.md
@@ -53,6 +53,8 @@ resource "twingate_connector" "aws_connector" {
 
 Import is supported using the following syntax:
 
+The [`terraform import` command](https://developer.hashicorp.com/terraform/cli/commands/import) can be used, for example:
+
 ```shell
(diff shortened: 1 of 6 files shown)


twingate-blee (approved):
lgtm

PR #960, opened 2026-09-08
fix: Stop sending an all-inclusive type filter in the groups query

ekampf:
## What

`NewGroupFilterInput` seeded **every** `twingate_groups` query with

```json
"type": {"in": ["MANUAL", "SYNCED", "SYSTEM"]}
```

even when the config never mentioned `types`. So this:

```hcl
data "twingate_groups" "foo" {
  name_contains = "foo"
}
```

sent a type filter the user never asked for.

## Why it matters

`MANUAL`/`SYNCED`/`SYSTEM` is the complete set of group types the provider knows (`model/group.go:11-15` — the same three are the only values accepted by the `types` validator and by the provider-level cache filter). So the filter selects everything and is a no-op **today**.

The problem is that it's a hardcoded allow-list. The day the Twingate API grows a fourth group type, this silently turns into an *exclusion* filter and the provider stops returning groups it should — with no error and nothing in the schema hinting at why.

These defaults were not a deliberate product decision. They arrived in b1cb43a "Fix group filtering using API (#271)", which moved filtering from a client-side `GroupsFilter.Match` helper into the GraphQL `filter:` argument. The helper it replaced applied `Type`/`IsActive` **only when non-nil** — no defaults. No rationale in the commit message, no accompanying doc change.

## The change

`GroupFilterInput.Type` becomes an optional pointer with `omitempty`, allocated only when the caller sets `types`:

```go
Type *GroupTypeFilterOperatorInput `json:"type,omitempty"`
```

Payload before / after, for `name_contains = "foo"`:

```diff
-{"name":{...,"contains":"foo",...},"type":{"in":["MANUAL","SYNCED","SYSTEM"]},"isActive":{"eq":true}}
+{"name":{...,"contains":"foo",...},"isActive":{"eq":true}}
```

When `types` **is** set, nothing changes: `{"type":{"in":["MANUAL"]},...}`.

Note this is the first use of `omitempty` in the repo — the prevailing convention is pointer + nil → explicit `null`, as `name` still does. `omitempty` is deliberate here because the goal is the key's *absence*, not a `null`.

## Not in scope

The sibling `isActive: {eq: true}` default is **left exactly as it is**. Removing it would be user-visible — `data "twingate_groups" { name = "x" }` would start returning inactive groups — so it belongs in its own appropriately-flagged change.

Two related inconsistencies turned up while investigating, both untouched here and worth a follow-up:

- `buildFilter` returns `nil` when no filter attribute is set, so `data "twingate_groups" "all" {}` sends no filter and *does* return inactive groups — while `{ name = "x" }` silently restricts to active. Same datasource, opposite semantics.
- The cache path (`model.Group.Match`) applies `isActive` only when explicitly set, so it already disagrees with the API path's default.

## Behavior change

None for the three known group types. A future server-side group type will no longer be silently excluded.

## Tests

- `TestBuildGroupsFilter` — updated; it asserts whole-struct equality, so every non-nil case needed its `Type` expectation adjusted.
- `TestBuildGroupsFilterSerialization` — **new**. Asserts on the marshaled JSON payload, which nothing previously covered (the client tests use `httpmock` responders and never read the request body). Checks that `type` is absent when unset, present with exactly the given values when set, and that `isActive` is still sent in both cases — that last one locks in the deliberate decision above, so a future change to it has to be conscious.

Verified locally: `go build ./...`, full `go test ./twingate/internal/...`, `gofmt` clean, and `tfplugindocs generate` produces no diff.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@
 .*.swp
 *.iml
 *.test
+*.tmp
 
 # terraform files
 terraform.tfvars
--- a/twingate/internal/client/query/groups-read.go
+++ b/twingate/internal/client/query/groups-read.go
@@ -31,9 +31,9 @@ func (u Groups) ToModel() []*model.Group {
 }
 
 type GroupFilterInput struct {
-	Name     *StringFilterOperationInput  `json:"name"`
-	Type     GroupTypeFilterOperatorInput `json:"type"`
-	IsActive BooleanFilterOperatorInput   `json:"isActive"`
+	Name     *StringFilterOperationInput   `json:"name"`
+	Type     *GroupTypeFilterOperatorInput `json:"type,omitempty"`
+	IsActive BooleanFilterOperatorInput    `json:"isActive"`
 }
 
 type StringFilterOperationInput struct {
@@ -92,15 +92,8 @@ func NewGroupFilterInput(input *model.GroupsFilter) *GroupFilterInput {
 		return nil
 	}
 
-	// default filter settings
+	// groups are filtered by active state unless the caller asks otherwise
 	filter := &GroupFilterInput{
-		Type: GroupTypeFilterOperatorInput{
-			In: []string{
-				model.GroupTypeManual,
-				model.GroupTypeSynced,
-				model.GroupTypeSystem,
-			},
-		},
 		IsActive: BooleanFilterOperatorInput{Eq: true},
 	}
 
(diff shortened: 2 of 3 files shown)


twingate-blee (changes requested):
Add  *.tmp to .gitignore and remove the file query_test.go.tmp

vmanilo (approved):

twingate-blee (approved):
lgtm

PR #970, opened 2026-09-09
ci: Serialize dependabot CI runs and batch dependabot PRs

ekampf:
## Problem

Acceptance tests run against a **single shared Twingate tenant** (`secrets.TWINGATE_URL` / `TWINGATE_NETWORK` / `TWINGATE_API_TOKEN`; `Makefile:8` pins `SWEEP_TENANT=terraformtests`). Concurrent CI runs corrupt each other through one job:

`cleanup-before-acctests` runs `make sweep-all-twingate_x509_certificate_authority` → `Makefile:39-40` → `Makefile:35` sets `SWEEP_DELETE_ALL=true`, which at `sweeper_test.go:95-101` **short-circuits the per-run prefix filter and deletes every `twingate_x509_certificate_authority` in the tenant, regardless of which run created it.** That job has no `needs:`, so it fires within a minute of a run starting.

So: PR-B is mid-acceptance-test → PR-A's run starts → PR-A's sweep deletes the x509 CAs PR-B is asserting on → PR-B fails with vanished-resource errors that look like a provider bug.

Everything *else* is properly namespaced — `TEST_UNIQUE_VALUE` embeds `github.run_id`, and the post-run `cleanup` sweeper filters on that run's own prefix, so two runs' cleanup jobs can't touch each other's resources. `SWEEP_DELETE_ALL` is the whole problem.

Meanwhile the concurrency group was keyed **per-PR**, so it deduped a PR against itself and did nothing across PRs.

## Why two files and not just `ci.yml`

**A concurrency group is not a queue.** It holds one in-progress run plus **one** pending run; queueing a third *cancels the pending one*.

Six dependabot PRs merged on 2026-09-08 (#951–#956, contiguous). With a shared group and nothing else, one would run, one would pend, and **the middle four would be cancelled** — leaving a failed `CI Finished` and a stalled auto-merge. So the burst has to be prevented, not just serialized.

`dependabot.yml` used **no `groups:` batching** — every dependency got its own PR — with an effective ceiling of **20 simultaneous PRs** (10 gomod + 5 + 5 from the two ecosystems that set no limit).

## Changes

**`.github/workflows/ci.yml`** — dependabot PRs share one group:

| trigger | group | `cancel-in-progress` |
|---|---|---|
| dependabot PR | `CI-dependabot` | `false` |
| human PR | `CI-<number>` | `true` |
| `push` to `main` | `CI-refs/heads/main` | `true` |

Human PRs and pushes behave exactly as before.

**`.github/dependabot.yml`** — batch minor/patch bumps per ecosystem, and set `open-pull-requests-limit: 3` on `github-actions` and `docker`.

## Two deliberate choices worth reviewing

**`update-types: [minor, patch]` on the groups is load-bearing.** `dependabot/fetch-metadata@v3` reports the *highest* semver bump across a grouped PR, and `automerge-dependabot.yaml:19` only auto-merges `semver-minor`/`semver-patch`. A group that swept in one major bump would report `semver-major` and **silently stop auto-merging the entire batch**. Majors stay individual, which matches the existing policy of reviewing them by hand.

**`github.event.pull_request.user.login`, not `github.actor`** (the idiom in `automerge-dependabot.yaml:11`). `github.actor` flips to a human if someone pushes a commit onto a dependabot branch — the PR still hits the shared tenant, so it must stay serialized — and is `dependabot[bot]` on the post-automerge push to `main`, which would wrongly pull main pushes into the dependabot group.

**Accepted cost:** `cancel-in-progress` is per-group, so it must be `false` for dependabot — `true` would let one dependabot PR cancel a *different* one's in-flight run. Consequence: a dependabot rebase force-push no longer cancels its own superseded run.

## Verification

`actionlint` output is **byte-identical to `main` apart from line-number shifts** (+2 from the longer comment) — same 13 findings, no new ones, verified by diffing against a pristine `origin/main` worktree. `dependabot.yml` parses and yields the intended structure.

The concurrency expression can't be evaluated locally; GitHub resolves it at queue time. After merge, on the next dependabot cycle: `gh run list --workflow=CI` should show one run `in_progress` with others `queued` rather than several `in_progress`, and no `cancelled` CI runs on dependabot PRs.

One thing I could not verify: with `directories:` (plural) plus `groups:`, it's unclear whether dependabot emits one PR spanning both `/` and `/tools/upgrader` or one per directory. Worst case gomod yields 2 grouped PRs instead of 1 — still far below the current 6, and the serialization covers it either way.

## Follow-ups (not fixed here)

1. **The actual root cause** — `SWEEP_DELETE_ALL=true` deleting x509 CAs tenant-wide. An age-scoped filter would make parallel runs non-destructive and render this serialization optional.
2. **`smoketests.yml` is unguarded** — it runs the same `sweep-all` x509 nuke on `cron: "0 */3 * * *"` with **no `concurrency:` block**. Since every group key starts with `${{ github.workflow }}`, CI and smoketests can never share a group, so a scheduled smoketest can still kill an in-flight CI run. Closing that needs a literal shared group name across both workflows.
3. **The connector sweeper is dead code** — `connector_test.go:20` filters on `test.Prefix()`, but connector names are `tf-acc-<sha224(prefix)[:15]>-<rand>` (`helper.go:18-27`), a *hash* of the prefix. It matches nothing, so connectors accumulate indefinitely.
4. **`CI Finished` never reports on fork PRs** — the `finish` job has no `if: always()`, so it's skipped when the acceptance/cleanup jobs are skipped on forks.


.github/dependabot.yml
```diff
@@ -10,11 +10,42 @@ updates:
   open-pull-requests-limit: 10
```
twingate-blee, line 10: We should probably also make this open-pull-requests-limit: 3

twingate-blee (approved):
see comment which is optional

PR #974, opened 2026-09-10
Fix: sweep x509 certs

vmanilo:
Tests passes: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/34503635051

## Changes
- removed cleanup step before each CI run
- set on demand action to clean up certs


twingate-blee (approved):
lgtm

PR #976, opened 2026-09-11
Renamed getTags to getKeyValueMap

### LINKED ISSUE #919: Rename `getTags()` to `getKeyValueMap()`
Following up on this [discussion](https://github.com/Twingate/terraform-provider-twingate/pull/905#discussion_r3808956579), we should rename `getTags()` to `getKeyValueMap()` and drop the `getHeaderRewrites()` wrapper that exists only to explain the mismatch. `getTags` is a generic `types.Map` → `map[string]string` converter.

Nothing in `getTags()` is tag-specific, it collapses a null, unknown or empty map to `nil` and otherwise copies string values:

```go
func getTags(rawTags types.Map) map[string]string {
	if rawTags.IsNull() || rawTags.IsUnknown() || len(rawTags.Elements()) == 0 {
		return nil
	}
	...
}
```

The web app resource reuses it for `request_header_rewrites`, and because the name says "tags", that reuse needs a wrapper plus three lines of comments to justify itself:

```go
// getHeaderRewrites converts the configured map for the API. Header rewrites and
// tags are both plain string maps, so the tag converter applies: a null, unknown
// or empty map becomes nil, which the client sends as an empty list.
func getHeaderRewrites(rawRewrites types.Map) map[string]string {
	return getTags(rawRewrites)
}
```

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/919

Test runs [UPD]: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/34687211530

## Changes
- code refactoring


twingate-blee:
When testing with
```
resource "twingate_ssh_resource" "tag_test" {
  name              = "SSH Bastion"
  address           = "10.128.0.47"
  remote_network_id = twingate_remote_network.gcp.id

  tags = {}          # <-- explicitly empty
}
```

It has the error

╷
│ Error: Provider produced inconsistent result after apply
│
│ When applying changes to twingate_ssh_resource.tag_test, provider "provider[\"registry.terraform.io/twingate/twingate\"]" produced an unexpected new value: .tags: was cty.MapValEmpty(cty.String), but now null.
│
│ This is a bug in the provider, which should be reported in the provider's own issue tracker.

twingate-blee (approved):
lgtm

clement0010 (approved):
LGTM 👍🏻

PR #977, opened 2026-09-13
Fix: removed unreachable nil check

### LINKED ISSUE #918: Remove unreachable `nil, nil` branches in the SSH and Kubernetes resource clients
Follow up from this [discussion](https://github.com/Twingate/terraform-provider-twingate/pull/905#discussion_r3786426939), we should remove the `if res == nil { return nil, nil //nolint:nilnil }` branches after `client.mutate` in `ssh-resource.go` and `kubernetes-resource.go`. This is because are unreachable and should be dropped.

`client.mutate` already fails before `ToModel()` is ever reached when the mutation payload has no entity:

```go
// twingate/internal/client/client.go:352-354
if resp.IsEmpty() {
    return opr.apiError(ErrGraphqlResultIsEmpty, attrs...)
}
```

For every create/update response type, `IsEmpty()` and the nil branch of `ToModel()` test the *same* condition — `q.Entity == nil` (e.g. `query/ssh-resource-create.go`, `query/kubernetes-resource-update.go`). So once `mutate` returns without an error, `response.ToModel()` cannot be `nil`.

vmanilo:
Resolves https://github.com/Twingate/terraform-provider-twingate/issues/918

Test runs: https://github.com/vmanilo/terraform-provider-twingate/actions/runs/34779160773

## Changes
- removed unreachable nil check


twingate-blee (approved):
lgtm

clement0010 (approved):
LGTM 👍🏻

